PR #2, opened 2023-06-21
CSS-4560: Add initial worker functionality

kelkawi-a:
This PR adds initial Temporal worker functionality. This entails the following:
- Introduction of an OCI image pre-installed with Python to run the Temporal worker.
- Reading sample workflow (currently stored inside the charm) which can be executed by the worker. This will later on be replaced by a user-provided file containing supported workflows and activities.
- Config necessary for Temporal worker and authentication (tested against our Candid-based deployment of Temporal).

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -12,3 +12,111 @@ options:
       Acceptable values are: "info", "debug", "warning", "error" and "critical"
     default: "info"
     type: string
+
+  host:
+    description: The hostname of the Temporal server.
+    default: ""
+    type: string
+
+  namespace:
+    description: Temporal namespace the worker should connect to.
+    default: ""
+    type: string
+
+  queue:
+    description: Temporal task queue the worker should connect to
+    default: ""
+    type: string
+
+  encryption-key:
+    description: Base64-encoded key used for data encryption.
+    default: ""
+    type: string
+
+  auth-enabled:
+      description: Determines if authentication is enabled or not.
+      default: False
+      type: boolean    
+
+  auth-provider:
+    description: |
+      Authentication provider for user authentication.
+    
+      Acceptable values are "candid" and "google".
+    default: ""
+    type: string
+
+  tls-root-cas:
+    description: Root certificate authority (CA) certificates for TLS communication.
+    default: ""
+    type: string
+
+  candid-url:
+    description: URL of the Candid service.
+    default: ""
+    type: string
+
+  candid-username:
+    description: Username for accessing the Candid service.
+    default: ""
+    type: string
+
+  candid-public-key:
+    description: Public key for Candid authentication.
+    default: ""
+    type: string
+
+  candid-private-key:
+    description: Private key for Candid authentication.
+    default: ""
+    type: string
+
+  oidc-auth-type:
+    description: Authentication type for OIDC authentication.
+    default: "service_account"
+    type: string
+
+  oidc-project-id:
+    description: Project ID for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-private-key-id:
+    description: Private key ID for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-private-key:
+    description: Project key for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-client-email:
+    description: Client email for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-client-id:
+    description: Client ID for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-auth-uri:
+    description: Authentication URI for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-token-uri:
+    description: Token URI for OIDC authentication.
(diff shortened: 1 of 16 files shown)


src/charm.py
```diff
@@ -1,103 +1,202 @@
...
+        """Check if required config has been set by user.
+
+        Args:
+            config_list: list of required config parameters.
+
+        Raises:
+            ValueError: if any of the required config is not set.
+        """
+        for param in config_list:
+            if self.config[param].strip() == "":
```
mertalpt, line 108: Is there a possibility that the key is missing from the received configuration? I am not too well-versed in charms. If so, would it be a good idea to catch the exception to re-emit it better formatted?

src/charm.py
```diff
@@ -1,103 +1,202 @@
...
+        "/temporal_client/activities.py",
+        resources_path / "temporal_client/activities.py",
+    )
+
+    # Install worker dependencies
+    worker_dependencies_path = resources_path / "worker-dependencies.txt"
+    with open(worker_dependencies_path, "r") as dependencies_file:
+        dependencies = dependencies_file.read().split("\n")
+        logger.info(f"installing worker dependencies {dependencies}...")
+        container.exec(["pip", "install", *dependencies]).wait()
```
mertalpt, line 183: I think this is fine. Just wondering if you considered `pip install -r {dependencies}`. Also, do we need validation? What happens if someone puts a line like `; rm -rf /` in the dependency file?

mertalpt (commented):
I think there is discussion to be had, not approving yet.

src/charm.py
kelkawi-a: For now, these are the necessary Temporal dependencies bundled with the charm. When we add an action for allowing a user to install dependencies on the workload container, I will make sure to add validation.

src/charm.py
kelkawi-a: In juju, you cannot remove a config, you can only unset it (sending it back to its default value). So there will be no need to check if the key is missing from the configuration.

config.yaml
```diff
@@ -12,3 +12,111 @@ options:
...
+    type: string
+
+  queue:
+    description: Temporal task queue the worker should connect to
+    default: ""
+    type: string
+
+  encryption-key:
+    description: Base64-encoded key used for data encryption.
+    default: ""
```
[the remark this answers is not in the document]
kian99: This should ideally be using juju secrets if this charm is only designed to support juju 3 and up, if it needs to support juju 2.9 then it will have to live here like this.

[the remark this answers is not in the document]
kelkawi-a: You're right, looking at some other charms (e.g. [postgresql-k8s](https://charmhub.io/postgresql-k8s)) it seems like the practice is to include sensitive data such as passwords as part of the charm's data bucket through peer relation. As this is not a published charm yet, I will keep it as is and refactor it in the coming PRs.

@kian99 any other opinions on this?

config.yaml
```diff
@@ -12,3 +12,111 @@ options:
...
+    description: Token URI for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-auth-cert-url:
+    description: Authentication certificate URL for OIDC authentication.
+    default: ""
+    type: string
+
+  oidc-client-cert-url:
```
kian99, line 119: Just a comment, the config here is getting a bit lengthy. Sometimes inevitable but would be good to eliminate where possible.

src/charm.py
kian99: +1 on validation

kian99 (approved):

config.yaml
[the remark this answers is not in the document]
kian99: I've seen the postgres-k8s charm accept secret config through actions which are then stored in the peer relation data buckets. I don't particularly like that approach because it requires manual setup after a deployment. But it is a possibility. My other comment above around whether you support juju 2.9 or 3 will also help lead your decision.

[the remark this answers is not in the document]
kelkawi-a: Similar to the other Temporal charms, this one will support juju 3. Regardless of the approach, will we not require a manual step if we are not using config? (i.e. an action that creates a juju secret, until user secrets are made available).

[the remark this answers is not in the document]
kelkawi-a: After a discussion with members of the Charmhub community, it seems that having these configuration secrets in config is the way to go for now until juju user secrets are implemented in juju 3.3 by the end of this cycle.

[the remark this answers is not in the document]
mertalpt: Good enough for me.

src/charm.py
mertalpt: Understood.

src/charm.py
mertalpt: Good enough for me.

mertalpt (approved):
All of my discussions are resolved. I think it would be a good idea to create issues for the future items from the discussions. @kelkawi-a

PR #3, opened 2023-06-26
CSS 4560: Add wheel file resource

kelkawi-a:
This PR introduces the "workflows-file" resource which can be attached by the user to the charm. This wheel file is then extracted by the charm code to extract Temporal workflow and activity definitions, and then starts the Temporal worker with the workflows and activities specified by the user (i.e. user specifies "GreetingWorkflow" as a supported workflow, the worker is started and listens for "GreetingWorkflow" triggers to be processed. The same applies to activities).

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,52 @@ juju deploy temporal-worker-k8s
 juju config temporal-worker-k8s --file=path/to/config.yaml
 ```
 
+### Attaching "workflows-file" resource
+
+The Temporal worker operator expects a "workflows-file" resource to be attached
+after deployment, which contains a set of defined Temporal workflows and
+activities as defined in the [resource_sample](./resource_sample/) directory.
+The structure of the built wheel file must follow the same structure:
+
+```
+- workflows/
+    - workflow-a.py
+    - workflow-b.py
+- activities/
+    - activity-a.py
+    - activity-b.py
+- some_other_direcory/
+- some_helper_file.py
+```
+
+The wheel file can be built by running `poetry build -f wheel` in the
+[resource_sample](./resource_sample/) directory.
+
+Once ready, the resource can be attached as follows:
+
+```bash
+juju attach-resource temporal-worker-k8s workflows-file=<path_to_wheel_file>
+```
+
+### Defining supported workflows and activities
+
+Once the wheel file is processed by the worker to extract all the workflows and
+activities, a list of supported workflows and activities must be defined by the
+user before the charm can be started. This can be done as follows:
+
+```bash
+juju run temporal-worker-k8s/0 add-workflows workflows="GreetingWorkflow"
+juju run temporal-worker-k8s/0 add-activities activities="compose_greeting"
+```
+
+Once done, the charm should enter an active state, indicating that the worker is
+running successfully. To verify this, you can check the logs of the kubernetes
+pod to ensure there are no errors with the workload container:
+
+```bash
+kubectl -n <juju_model_name> logs temporal-worker-k8s-0 -c temporal-worker -f
+```
+
 ## Verifying
 
 To verify that the setup is running correctly, run `juju status --watch 1s` and
--- a/src/actions/activities.py
+++ b/src/actions/activities.py
@@ -1,0 +1,104 @@
+# Copyright 2023 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+"""Define the Temporal worker activities actions."""
+
+# pylint: disable=W0212,R0801
+
+import logging
+
+from ops import framework
+
+from log import log_event_handler
+
+logger = logging.getLogger(__name__)
+
+
+class ActivitiesActions(framework.Object):
+    """Client for Temporal worker activities actions."""
+
+    def __init__(self, charm):
+        """Construct.
+
+        Args:
+            charm: The charm to attach the hooks to.
+        """
+        super().__init__(charm, "activities-actions")
+        self.charm = charm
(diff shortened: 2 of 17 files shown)


README.md
```diff
@@ -25,6 +25,52 @@ juju deploy temporal-worker-k8s
...
+
+```
+- workflows/
+    - workflow-a.py
+    - workflow-b.py
+- activities/
+    - activity-a.py
+    - activity-b.py
+- some_other_direcory/
+- some_helper_file.py
```
kian99, line 43: Does `some_helper_file.py` need to be in `some_other_directory`?

README.md
```diff
@@ -25,6 +25,52 @@ juju deploy temporal-worker-k8s
...
+```
+
+### Defining supported workflows and activities
+
+Once the wheel file is processed by the worker to extract all the workflows and
+activities, a list of supported workflows and activities must be defined by the
+user before the charm can be started. This can be done as follows:
+
+```bash
+juju run temporal-worker-k8s/0 add-workflows workflows="GreetingWorkflow"
```
kian99, line 62: Would it be beneficial to provide an `all` value?

src/actions/activities.py
```diff
@@ -0,0 +1,104 @@
...
+        if not self.charm._state.is_ready():
+            event.defer()
+            return
+
+        if not self.charm.unit.is_leader():
+            event.fail("action cannot be performed on non-leader unit")
+            return
+
+        if self.charm._state.supported_activities is None:
+            self.charm._state.supported_activities = []
```
kian99, line 77: Maybe these should be set as defaults on the state? Though I guess they could be deleted elsewhere so probably best to keep the check.

src/charm.py
```diff
@@ -46,11 +39,15 @@ def __init__(self, *args):
             args: Ignore.
         """
         super().__init__(*args)
+        self._state = State(self.app, lambda: self.model.get_relation("peer"))
         self.name = "temporal-worker"
 
         self.framework.observe(self.on.config_changed, self._on_config_changed)
         self.framework.observe(self.on.temporal_worker_pebble_ready, self._on_temporal_worker_pebble_ready)
 
+        self.workflows_actions = WorkflowsActions(self)
```
kian99, line 48: This is neat

src/charm.py
```diff
@@ -95,6 +106,54 @@ def _validate(self):
...
+
+                # Push wheel file to the container and extract it.
+                container.push("/user_provided/wheel_file.whl", wheel_data, make_dirs=True)
+                container.exec(["apt-get", "update"]).wait_output()
+                container.exec(["apt-get", "install", "unzip"]).wait_output()
+                container.exec(["unzip", "/user_provided/wheel_file.whl", "-d", "/user_provided"]).wait_output()
+
+                # Find the name of the module provided by the user and set it in state.
+                command = "find /user_provided -mindepth 1 -maxdepth 1 -type d ! -name *.dist-info ! -name *.whl"
+                out, err = container.exec(command.split(" ")).wait_output()
```
kian99, line 147: We should probably handle the case where these execs fail for whatever reason.

kian99 (approved):
Impressive!

README.md
kelkawi-a: Not necessarily, just a way of showing that the user can have any folder structure they want as long as they have the mandatory ```workflows``` and ```activities``` directories.

README.md
kelkawi-a: Interesting, yes I believe it would be beneficial. Will keep it in mind in the next PR when I add a check for the ```@workflow.defn``` and ```@activity.defn``` decorators as well.

src/charm.py
kelkawi-a: Thanks :)

src/charm.py

src/actions/activities.py
kelkawi-a: Updated, thanks!

README.md
```diff
@@ -25,6 +25,52 @@ juju deploy temporal-worker-k8s
...
+The structure of the built wheel file must follow the same structure:
+
+```
+- workflows/
+    - workflow-a.py
+    - workflow-b.py
+- activities/
+    - activity-a.py
+    - activity-b.py
+- some_other_direcory/
```
mertalpt, line 42: Typo `direcory`.

src/actions/activities.py
```diff
@@ -0,0 +1,95 @@
...
+        if not self.charm._state.is_ready():
+            event.defer()
+            return
+
+        if not self.charm.unit.is_leader():
+            event.fail("action cannot be performed on non-leader unit")
+            return
+
+        sw = self.charm._state.supported_activities
+        for workflow in event.params["activities"].split(","):
```
mertalpt, line 48: Is there a chance that there is confusion from someone putting in `a, b, c` rather than `a,b,c` here? I did not see if you validate or strip this somewhere else.

There is similar usage in other places as well.

PS. After looking through the entire PR I guess you use `split(",")` everywhere so you would not have problems with strings not matching. Still, if this string ever faces the user (e.g. in the logs) it would be nicer to trim the whitespace, What do you think?

mertalpt (approved):
LGTM.

README.md
kelkawi-a: Updated, thanks!

src/actions/activities.py
kelkawi-a: You're right, i'll trim the whitespace and add some minor validation to the user input in the next PR when I add the other charm actions.

PR #4, opened 2023-06-27
CSS 4560: Add .env resource file and dependency actions

kelkawi-a:
This PR addresses the following:
- Adds an "env-file" optional resource which the user can attach to the charm. This is a .env file which contains environment variables to be injected into the workload container to be used by the worker.
- Adds a config value for setting the ```workflows-file-name``` resource. This allows us to rename the resource file that is attached to the charm, and ```pip install``` it and its dependencies onto the workload container.
- Add minor validation to user input for workflow and activity actions.


kelkawi-a:
> A few more options would be:
> 
> 1. have the wheel name as config (i guess it is simpler for them to write again the name of the wheel rather than dependencies)
> 2. reconstructing the wheel name based on the wheel (see below a code example)
> 3. telling juju team to use the default name of the resource if none is given (probably the cleanest)
> 
> Personally, I would have started the convo with the juju team (3), and meanwhile I would have done (1).. if wheel name config is not there, then you try to guess it using (2), if that fails then put the charm in a state where it asks the user to config the wheel name
> 
> potential code for reconstructing the name
> 
> ```python
> default_wheel_name = os.path.basename(resource_path)
> 
> # get the base name from the dist-info folder
> command = "find /user_provided -type d -name *.dist-info"
> dist_info, error = container.exec(command.split(" ")).wait_output()
> # check error
> dist_info = dist_info.strip()
> 
> # get the tag name from the *.dist-info/WHEEL file
> command = ["awk", "-F", ": ", "/^Tag:/ {print $2}", f"{dist_info}/WHEEL"]
> wheel_info, error = container.exec(command).wait_output()
> # check error
> 
> # merge them
> wheel_base = os.path.basename(dist_info).replace(".dist-info", "")
> wheel_tag = wheel_info.strip()
> original_wheel_name = f'{wheel_base}-{wheel_tag}.whl'    
> 
> # rename wheel to the original name
> command = f"mv /user_provided/{default_wheel_name} /user_provided/{original_wheel_name}"
> container.exec(command.split(' ')).wait_output()
> 
> # and install the wheel here:
> command = f"pip install /user_provided/{original_wheel_name}"
> container.exec(command.split(' ')).wait_output()
> ```
> 
> This code is kinda weak cause not all wheels have the same structure, but modern ones tend to have a dist-info

I fully agree. I had thought of using config for accessing the wheel file name but steered away from it eventually, but I do believe it's a better alternative. I've refactored the code to allow the user to set the wheel file name through config (with some validation on the file name), and used ```pip install``` to install all its dependencies onto the workload container.

As this should be considered a temporary solution, I've submitted a [bug report](https://bugs.launchpad.net/juju/+bug/2025601) for the juju folks regarding this issue to prevent the renaming of the resource file that gets attached.

Thank you!

src/charm.py
```diff
@@ -183,14 +182,57 @@ def _check_required_config(self, config_list):
...
+            if not self.config["auth-provider"]:
+                raise ValueError("Invalid config: auth-provider value missing")
+
+            if not self.config["auth-provider"] in SUPPORTED_AUTH_PROVIDERS:
+                raise ValueError("Invalid config: auth-provider not supported")
+
+            if self.config["auth-provider"] == "candid":
+                self._check_required_config(REQUIRED_CANDID_CONFIG)
+
+            if self.config["auth-provider"] == "google":
```
kian99, line 225: Could be an if else here? Or a switch-case, I know that a thing from a fairly recent version of Python.

kian99 (approved):
Just from a readability and code point of view lgtm. I think the approach with config for filenames is sound.

gtato (approved):

src/charm.py
```diff
@@ -151,21 +137,34 @@ def _process_wheel_file(self, event):
 
                 # Push wheel file to the container and extract it.
                 container.push("/user_provided/wheel_file.whl", wheel_data, make_dirs=True)
```
gtato, line 139: maybe you would want to remove the hardcoded wheel_file.whl from here and get it from resource_path

src/charm.py
```diff
@@ -151,21 +137,34 @@ def _process_wheel_file(self, event):
...
+                    logger.error(f"failed to extract module name from wheel file: {error}")
                     raise ValueError("Invalid state: failed to extract module name from wheel file")
 
                 module_name = out.split("\n")
                 if self.unit.is_leader():
                     self._state.module_name = module_name[0].split("/")[-1]
 
+                # Rename wheel file to its original name and install it
+                container.exec(
+                    ["mv", "/user_provided/wheel_file.whl", f"/user_provided/{self.config['workflows-file-name']}"]
```
gtato, line 158: same here

src/charm.py
kelkawi-a: Noted, will make this change in the next PR along with some minor changes.

PR #6, opened 2023-07-13
Update "publish_charm" workflow

kelkawi-a:


src/charm.py
```diff
@@ -123,6 +129,7 @@ def _process_wheel_file(self, event):  # noqa: C901
 
         try:
             resource_path = self.model.resources.fetch("workflows-file")
+            filename = str(resource_path).rsplit("/", maxsplit=1)
```
zorzigio, line 132: There might be a problem here. `rsplit` returns a list. Did I miss something?
A better way of doing this might be to use `pathlib`

```python
from pathlib import Path
filename = Path(resource_path).name
```

src/charm.py
```diff
@@ -136,10 +143,10 @@ def _process_wheel_file(self, event):  # noqa: C901
                 wheel_data = file.read()
 
                 # Push wheel file to the container and extract it.
-                container.push("/user_provided/wheel_file.whl", wheel_data, make_dirs=True)
+                container.push(f"/user_provided/{filename}", wheel_data, make_dirs=True)
                 container.exec(["apt-get", "update"]).wait()
                 container.exec(["apt-get", "install", "unzip"]).wait()
```
zorzigio, line 148: Would it not be better to bake the `apt-get` logic directly into the charm (if possible), or put it in the `__init__` function so you don't have to do it every time you update the wheel?

src/charm.py
```diff
@@ -151,11 +158,11 @@ def _process_wheel_file(self, event):  # noqa: C901
 
                 module_name = out.split("\n")
                 if self.unit.is_leader():
-                    self._state.module_name = module_name[0].split("/")[-1]
+                    self._state.module_name = module_name[0].rsplit("/", maxsplit=1)
```
zorzigio, line 161: As previous comment, `rsplit` returns a list.
You could use `pathlib` for this as well.
Also, would it make sense to do some error handling on `module_name`?

src/charm.py
zorzigio, line 1: F841 on line 102 and 175 are not needed

zorzigio (commented):
Great work @kelkawi-a. Can't wait to use the worker!

A general question: would it not be easier for the user to get activities and workflows added automatically when importing the wheel rather than having to add them all manually? Which would become tedious in case there are many workflows and activities

src/charm.py
kian99: I think the wording of this comment is confusing, do you mean put the apt-get logic into the workload container? I.e. Add it into the original docker image? If so, +1 to that, seems like you'll always need unzip so just include it by default.

tox.ini
```diff
@@ -51,16 +51,16 @@ deps =
     flake8-docstrings-complete>=1.0.3
     flake8-test-docs>=1.0
 commands =
-    pydocstyle {[vars]src_path}
+    pydocstyle {[vars]all_path}
     codespell {toxinidir} --skip {toxinidir}/.git --skip {toxinidir}/.tox \
```
kian99, line 55: I remove codespell from my CI, feels very unnecessary.

kian99 (approved):
lgtm barring suggestions from Giorgio on using Path

src/charm.py

src/charm.py
kelkawi-a: As this charm allows users to deploy with their own image (which should include Python as a minimum), this approach ensures that ```unzip``` is always installed even if it is missing from the image. The logic is included in this method because it ensure that the pebble API is ready, this cannot be guaranteed in the ```__init__``` function (have also tried including it in the ```pebble_ready``` hook, but that did not work either).

tox.ini
kelkawi-a: Codespell is one of the tools necessary for the [operator workflows](https://github.com/canonical/is-charms-contributing-guide/blob/main/CONTRIBUTING.md#static-code-analysis) to function properly.

src/charm.py
kelkawi-a: It probably isn't necessary as the wheel file would not have built properly without a valid module name.

src/charm.py
```diff
@@ -88,12 +98,18 @@ def _process_env_file(self, event):
         Args:
             event: The event triggered when the relation changed.
         """
-        try:
+        if not self._state.is_ready():
+            event.defer()
+            return
+
+        if self.unit.is_leader():
             self._state.env = None
```
mertalpt, line 106: What happens if you get a `ModelError` but you already set `env` to `None`?

src/charm.py
```diff
@@ -136,10 +153,10 @@ def _process_wheel_file(self, event):  # noqa: C901
                 wheel_data = file.read()
 
                 # Push wheel file to the container and extract it.
-                container.push("/user_provided/wheel_file.whl", wheel_data, make_dirs=True)
+                container.push(f"/user_provided/{filename}", wheel_data, make_dirs=True)
                 container.exec(["apt-get", "update"]).wait()
                 container.exec(["apt-get", "install", "unzip"]).wait()
-                container.exec(["unzip", "/user_provided/wheel_file.whl", "-d", "/user_provided"]).wait()
+                container.exec(["unzip", f"/user_provided/{filename}", "-d", "/user_provided"]).wait()
```
mertalpt, line 159: If you are going to reuse paths it might be better to put it in a variable.

Later Edit: It looks like you reuse this a lot below.

src/resources/worker.py
```diff
@@ -78,9 +79,18 @@ def _import_modules(module_type, module_name, supported_modules):
         module_name = file_name[:-3]
         module = import_module(module_name)
 
-        for sm in supported_modules:
-            if hasattr(module, sm):
-                module_list.append(getattr(module, sm))
+        if "all" in supported_modules:
+            for _, obj in inspect.getmembers(module):
+                if module_type == "workflows":
+                    if inspect.isclass(obj) and inspect.getmodule(obj) is module:
```
gtato, line 85: bs comment: maybe you could add a 
```
if inspect.getmodule(obj) is not module:
    continue
is_workflow = module_type == "workflows" and inspect.isclass(obj)
is_activity = inspect.isfunction(obj)
if is_workflow or is_activity:
    module_list.append(obj)
```

src/charm.py
kelkawi-a: ```ModelError``` happens when the resource has not been attached yet, so this error should not happen if a user has attached a new resource.

PR #7, opened 2023-07-27
Revert to older version of operator workflows

kelkawi-a:
Some updates to the [operator workflows](https://github.com/canonical/operator-workflows) are meant to reuse charms built during integration tests when publishing the charm to the edge channel on Charmhub after merging with main branch. As we are skipping the integration test, the workflows currently have no fallback. We will opt to use an older version of the workflows until this part of the workflow is more stable.


AmberCharitos (approved):
LGTM

src/charm.py
```diff
@@ -152,11 +152,12 @@ def _process_wheel_file(self, event):  # noqa: C901
             with open(resource_path, "rb") as file:
                 wheel_data = file.read()
 
+                wheel_file = f"/user_provided/{filename}"
```
gtato, line 155:
```python
original_wheel_file = "/user_provided/{self.config['workflows-file-name']}"
```

gtato (approved):

PR #9, opened 2023-08-06
CSS-5032: Use Temporal lib worker

kelkawi-a:
- Changed the Temporal worker to use the [Temporal lib worker](https://github.com/canonical/temporal-lib-py) which now has a built-in Sentry interceptor. Added Sentry-specific config to the charm to be able to use it. This has been tested with some sample workflows. Sentry logs uncaught exceptions as well as when using ```logging.error```.
- Added restart action for the Temporal worker.


src/resources/sentry_interceptor.py
```diff
@@ -0,0 +1,91 @@
...
+            set_tag("temporal.activity.task_queue", activity_info.task_queue)
+            set_tag("temporal.workflow.namespace", activity_info.workflow_namespace)
+            set_tag("temporal.workflow.run_id", activity_info.workflow_run_id)
+            try:
+                return await super().execute_activity(input)
+            except Exception as e:
+                if len(input.args) == 1 and is_dataclass(input.args[0]):
+                    set_context("temporal.activity.input", asdict(input.args[0]))
+                set_context("temporal.activity.info", activity.info().__dict__)
+                capture_exception()
```
mertalpt, line 52: Sentry is supposed to capture uncaught exceptions by itself, do you intend to report caught exceptions as well or am I missing something?

README.md
```diff
@@ -102,6 +112,16 @@ To add more replicas you can use the juju scale-application functionality i.e.
...
 ```
 
+## Error Monitoring
+
+The Temporal worker operator has a built-in Sentry interceptor which can be used
+to intercept and capture errors from the Temporal SDK. To enable it, run the
+following command:
+
+```bash
+juju config temporal-worker-k8s sentry-dsn=<YOUR_SENTRY_DSN>
```
mertalpt, line 122: There is an envvar for Sentry `SENTRY_RELEASE` IIRC, that lets you version your releases so that Sentry can track lineage of errors between versions. It is not mandatory but it might be a nice functionality here.

src/resources/sentry_interceptor.py
```diff
@@ -0,0 +1,91 @@
...
+            try:
+                return await super().execute_workflow(input)
+            except Exception as e:
+                if len(input.args) == 1 and is_dataclass(input.args[0]):
+                    set_context("temporal.workflow.input", asdict(input.args[0]))
+                set_context("temporal.workflow.info", workflow.info().__dict__)
+
+                if not workflow.unsafe.is_replaying():
+                    with workflow.unsafe.sandbox_unrestricted():
+                        capture_exception()
```
mertalpt, line 76: Similar thing here.

mertalpt (commented):
I do not think I know Temporal enough to give a definitive review on this but I am tending towards LGTM. I am only commenting for now but if this drags on too long let me know and I will take another look.

src/resources/sentry_interceptor.py
kelkawi-a: This interceptor will capture uncaught exceptions, and will catch caught exceptions if ```logging.error()``` is [called](https://stackoverflow.com/questions/73456044/why-does-sentry-log-errors-caught-in-except-block).

README.md
kelkawi-a: Updated to include Sentry release and environment configurations. Thanks!

mertalpt (approved):

gtato:
did you actually test this with sentry?

src/resources/worker.py
```diff
@@ -120,7 +122,17 @@ async def run_worker(charm_config, supported_workflows, supported_activities, mo
         client_config.auth = AuthOptions(provider=charm_config["auth-provider"], config=_get_auth_header(charm_config))
 
     if charm_config["encryption-key"].strip() != "":
-        client_config.encryption = EncryptionOptions(key=charm_config["encryption-key"])
+        client_config.encryption = EncryptionOptions(key=charm_config["encryption-key"], compress=True)
+
+    interceptors = []
```
gtato, line 127: wondering if this should be part of the lib, and not this

kelkawi-a:
> did you actually test this with sentry?

Yes, uncaught exceptions are reported in Sentry as well as when ```logging.error``` is used.

src/resources/worker.py
kelkawi-a: The interceptor for the client is different to the intereceptor for the worker, activity interceptors are implemented at the worker-level.

kelkawi-a: After our conversation, I've updated our Temporal lib to also export a Worker which has a configurable Sentry interceptor. The charm now uses this worker.

gtato (approved):
well done!

PR #11, opened 2023-08-23
Use wheel unpack instead of unzip

kelkawi-a:
Usage of ```unzip``` requires it to be installed using ```apt-get``` which sometimes faces proxy issues. This change opts for the use of ```wheel unpack``` with some slight modifications to execute the same logic.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -147,6 +147,7 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
 
         if self.unit.is_leader():
             self._state.module_name = None
+            self._state.unpacked_file_name = None
 
         if self.config["workflows-file-name"].strip() == "":
             raise ValueError("Invalid config: wheel-file-name missing")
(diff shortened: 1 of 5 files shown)


src/charm.py
```diff
@@ -174,12 +174,29 @@ def _process_wheel_file(self, event):  # noqa: C901
...
+                # Rename wheel file to its original name and install it
+                container.exec(["mv", wheel_file, original_wheel_file]).wait()
+
+                _, error = container.exec(["pip", "install", original_wheel_file]).wait_output()
+                if error is not None and error.strip() != "" and not error.strip().startswith("WARNING"):
+                    logger.error(f"failed to install wheel file: {error}")
+                    raise ValueError("Invalid state: failed to install wheel file")
+
+                # Unpack wheel file
+                container.exec(["pip", "install", "wheel"]).wait()
```
gtato, line 187: can you add this at the worker dependencies ?

src/charm.py
```diff
@@ -174,12 +174,29 @@ def _process_wheel_file(self, event):  # noqa: C901
 
                 # Push wheel file to the container and extract it.
                 container.push(wheel_file, wheel_data, make_dirs=True)
-                container.exec(["apt-get", "update"]).wait()
-                container.exec(["apt-get", "install", "unzip"]).wait()
-                container.exec(["unzip", wheel_file, "-d", "/user_provided"]).wait()
+
```
gtato, line 177: nothing to do with this line, but maybe you don't have to be in the with block the whole time no?

src/charm.py

gtato (approved):
nice

PR #13, opened 2023-09-04
Refactor and add integration tests

kelkawi-a:
Juju's python library now has the ability to attach local resources to deployed charms, which allows us to test this charm locally. This PR introduces the following:
- Refactoring of the charm logic to pass environment variables to the worker rather than the charm's config as an argument.
- Addition of charm, scaling and safe upgrades integration tests.


tests/integration/helpers.py
```diff
@@ -121,3 +100,88 @@ async def perform_temporal_integrations(ops_test: OpsTest):
...
+    assert len(ops_test.model.applications[app].units) == units
+
+
+async def setup_temporal_ecosystem(ops_test: OpsTest):
+    """Scale the application to the provided number and wait for idle.
+
+    Args:
+        ops_test: PyTest object.
+    """
+    await ops_test.model.deploy(APP_NAME_SERVER, channel="edge")
```
mertalpt, line 135: What does the `await` wait for here? For an API call to be made or for the charm to actually get deployed? If the latter, it might be possible to call the four functions first and then wait for them later. Which should parallelize the wait and speed up the test.

mertalpt (approved):

tests/integration/helpers.py
kelkawi-a: Updated to use `asyncio.gather` to group them together. Thanks!

src/charm.py
```diff
@@ -125,7 +127,13 @@ def _process_env_file(self, event):
         try:
             resource_path = self.model.resources.fetch("env-file")
             env = dotenv_values(resource_path)
-            self._state.env = env
+
+            for key, _ in env.items():
+                if key.startswith("TWC_"):
```
gtato, line 132: i guess you could prefix these ones too, so they don't collide

kelkawi-a: Environment variables provided by the user shouldn't be modified as the user will be expecting to fetch them by their original name within their code.

PR #16, opened 2024-03-28
Unpin 'protobuf' version and add better worker error logging.

mertalpt:
Changes:

1. `protobuf==3.2.0` conflicted with the dependencies of a package I needed for my workflow. I unpinned it. @kelkawi-a was there a reason you pinned it to that version? Should we pin it to something else that is compatible?
2. Error message itself in `worker_status.txt` was not good enough to debug my issues. I changed it to print the whole stack trace, so we get exact lines of code to look at.

PS. I did not want to deal with `tox`, fingers crossed that this passes the checks.

### DIFF
--- a/src/resources/worker-dependencies.txt
+++ b/src/resources/worker-dependencies.txt
@@ -1,3 +1,2 @@
 temporal-lib-py==1.1.3
-protobuf==3.20.0
 wheel==0.41.2
--- a/src/resources/worker.py
+++ b/src/resources/worker.py
@@ -8,8 +8,10 @@
 import asyncio
 import glob
 import inspect
+import logging
 import os
 import sys
+import traceback
 from importlib import import_module
 
 from temporallib.auth import (
@@ -21,6 +23,9 @@ from temporallib.auth import (
 from temporallib.client import Client, Options
 from temporallib.encryption import EncryptionOptions
 from temporallib.worker import SentryOptions, Worker, WorkerOptions
+
+
+logger = logging.getLogger(__name__)
 
 
 def _get_auth_header():
@@ -156,7 +161,8 @@ async def run_worker(unpacked_file_name, module_name):
     except Exception as e:
         # If an error occurs, write the error message to the status file
         with open("worker_status.txt", "w") as status_file:
-            status_file.write(f"Error: {e}")
+            logger.exception("Error in the workflow:")
+            traceback.print_exception(type(e), e, e.__traceback__, file=status_file)
 
 
 if __name__ == "__main__":  # pragma: nocover
--- a/src/resources/check_status.py
+++ b/src/resources/check_status.py
@@ -18,10 +18,10 @@ def check_worker_status():
             status = status_file.read().strip()
             logger.info(f"Async status: {status}")
 
-        if "Error" in status:
-            exit_code = 1
-        else:
+        if status.startswith("Success"):
             exit_code = 0
+        else:
+            exit_code = 1
     except FileNotFoundError:
         logger.error("Status file not found. Worker is not running.")
         exit_code = 1


src/resources/worker.py
```diff
@@ -156,7 +157,7 @@ async def run_worker(unpacked_file_name, module_name):
     except Exception as e:
         # If an error occurs, write the error message to the status file
         with open("worker_status.txt", "w") as status_file:
-            status_file.write(f"Error: {e}")
+            traceback.print_exception(type(e), e, e.__traceback__, file=status_file)
```
kelkawi-a, line 160: Thanks for this. The purpose of this was so that the charm can run a health check as seen [here](https://github.com/canonical/temporal-worker-k8s-operator/blob/5aaa90ac0dc0a2fb14f95f52bbe18bdd6d1e105c/src/charm.py#L343). Can we also log this trace so that these logs can be observed using `kubectl logs` or the juju equivalent?

src/resources/worker-dependencies.txt
```diff
@@ -1,3 +1,3 @@
 temporal-lib-py==1.1.3
-protobuf==3.20.0
+protobuf>=3.20.0
```
kelkawi-a, line 2: It's been quite a while so I don't recall the reason for pinning this. Regardless, it seems to pass the integration test so we'll take the update.

Come to think of it, what was the outcome of attaching your workflow to the charm when your workflow had a higher version of `protobuf` in the dependencies? Did it complain about the version mismatch?

Update: I've revisited the charm code, and I believe I know where the problem is. These worker dependencies are being installed **after** installing the wheel file dependencies, and therefore overriding whatever dependency versions you've introduced in your workflow.

It might work better if we move [this portion](https://github.com/canonical/temporal-worker-k8s-operator/blob/5aaa90ac0dc0a2fb14f95f52bbe18bdd6d1e105c/src/charm.py#L302C1-L309C63) of the `_update()` method to be executed before [this portion](https://github.com/canonical/temporal-worker-k8s-operator/blob/5aaa90ac0dc0a2fb14f95f52bbe18bdd6d1e105c/src/charm.py#L296C1-L300C19), such that the method would look like this:

```python
    def _update(self, event):
        container = self.unit.get_container(self.name)
        if not container.can_connect():
            event.defer()
            self.unit.status = WaitingStatus("waiting for pebble api")
            return

        # ensure the container is set up
        _setup_container(container, self.config["http-proxy"])

        try:
            self._validate(event)
        except ValueError as err:
            self.unit.status = BlockedStatus(str(err))
            return
```

I'm trying to think of whether this option has an advantage over changing dependency versions, or we might even make both changes. What do you think?

kelkawi-a (commented):
Thanks for the updates! Some food for thought on the worker dependency changes.

src/resources/worker-dependencies.txt
mertalpt: It did complain about version mismatch:
```
ubuntu@temporal-dev:~/Workspaces/cs-repos/cs-workflows/workflows/hrc_worklytics/dist$ juju attach-resource hrc-worklytics-worker workflows-file=./dist/hrc_worklytics-0.1.0-py3-none-any.whl
ERROR failed to upload resource "workflows-file": open ./dist/hrc_worklytics-0.1.0-py3-none-any.whl: no such file or directory
ubuntu@temporal-dev:~/Workspaces/cs-repos/cs-workflows/workflows/hrc_worklytics/dist$ juju attach-resource hrc-worklytics-worker workflows-file=./hrc_worklytics-0.1.0-py3-none-any.whl
ubuntu@temporal-dev:~/Workspaces/cs-repos/cs-workflows/workflows/hrc_worklytics/dist$ juju status --watch 2s
ubuntu@temporal-dev:~/Workspaces/cs-repos/cs-workflows/workflows/hrc_worklytics/dist$ juju debug-log
unit-hrc-worklytics-worker-0: 17:15:40 DEBUG unit.hrc-worklytics-worker/0.juju-log Emitting Juju event config_changed.
unit-hrc-worklytics-worker-0: 17:15:40 INFO unit.hrc-worklytics-worker/0.juju-log * running TemporalWorkerK8SOperatorCharm._on_config_changed
unit-hrc-worklytics-worker-0: 17:16:19 INFO unit.hrc-worklytics-worker/0.juju-log pushing /var/lib/juju/agents/unit-hrc-worklytics-worker-0/charm/src/resources/worker.py source...
unit-hrc-worklytics-worker-0: 17:16:20 INFO unit.hrc-worklytics-worker/0.juju-log pushing /var/lib/juju/agents/unit-hrc-worklytics-worker-0/charm/src/resources/check_status.py source...
unit-hrc-worklytics-worker-0: 17:16:20 INFO unit.hrc-worklytics-worker/0.juju-log pushing /var/lib/juju/agents/unit-hrc-worklytics-worker-0/charm/src/resources/worker-dependencies.txt source...
unit-hrc-worklytics-worker-0: 17:16:20 INFO unit.hrc-worklytics-worker/0.juju-log installing worker dependencies...
unit-hrc-worklytics-worker-0: 17:16:24 ERROR unit.hrc-worklytics-worker/0.juju-log failed to install worker dependencies: ERROR: googleapis-common-protos 1.63.0 has requirement protobuf!=3.20.0,!=3.20.1,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5,<5.0.0.dev0,>=3.19.5, but you'll have protobuf 3.20.0 which is incompatible.
ERROR: google-api-core 2.18.0 has requirement protobuf!=3.20.0,!=3.20.1,!=4.21.0,!=4.21.1,!=4.21.2,!=4.21.3,!=4.21.4,!=4.21.5,<5.0.0.dev0,>=3.19.5, but you'll have protobuf 3.20.0 which is incompatible.
WARNING: You are using pip version 20.1; however, version 24.0 is available.
You should consider upgrading via the '/usr/local/bin/python -m pip install --upgrade pip' command.
```

I think the worker itself should be able to impose its own restrictions on the dependencies and worker and the workflow should not override each other. On the other hand, if `protobuf` is a dependency of `temporallib`, it is `temporallib`s job to manage the version requirements. I don't think we need to specify `protobuf` here at all, it should be pulled as a dependency automatically.

TL;DR:
1. Worker and workflow should not override each other, we should try to merge the dependencies or maybe run the workflows in a virtual environment.
2. We should remove packages from requirements that are dependencies of dependencies. If certain versions do not work, specify them on the package who pulls that dependency.

src/resources/worker.py
mertalpt: Does this break ``check_status.py`?

I don't know the answer to your question. Worker code is not a part of the charm, is it?

src/resources/worker-dependencies.txt
kelkawi-a: Sounds fair. Shall we remove this dependency altogether and make it the workflow's responsibility to bring its required dependencies?

src/resources/worker.py
kelkawi-a: No it doesn't, the integration test seems to be working fine, this improves it as you said by adding the full trace rather than just the error.

I was asking if you could also add an additional line to log the error stack trace, which can then be observed on the workload container through `kubectl logs temporal-worker-k8s-0 -c temporal-worker`.

mertalpt:
@kelkawi-a After the approvals I would prefer to squash the commits before a rebase and merge. So it would be nice if you did not merge it yourself, unless you would squash and edit the commit messages yourself.

src/resources/worker.py
mertalpt: `logger.exception` is expected to print the traceback after the message.

src/resources/worker-dependencies.txt
mertalpt: Removed it.

kelkawi-a (approved):
LGTM, thank you!

gtato (approved):
lg

PR #18, opened 2024-06-06
Add vault relation

kelkawi-a:
This PR:

- Adds a relation to the [Vault operator](https://charmhub.io/vault-k8s) along with unit and integration tests
- Refactors unit tests file

Note: Code under the `lib` directory is fetched using `charmcraft fetch-lib` as a dependency and does not need to be reviewed.


requirements.txt
```diff
@@ -1,3 +1,4 @@
-ops >= 2.2.0
+ops==2.14.0
```
mertalpt, line 1: question (non-blocking): What is the reason for rolling this back? How long will it stay this way?

resource_sample/dist/python_samples-1.1.0-py3-none-any.whl
mertalpt, line 1: thought (non-blocking): If the intention is to provide a ready-to-use example, I wonder if it would be better to supply a `Makefile` instead. I don't think having opaque binaries in a repository is good.

mertalpt (approved):

README.md
```diff
@@ -137,14 +137,38 @@ juju relate temporal-worker-k8s admin/cos.loki
...
+## Vault
+
+The Temporal worker operator charm can be related to the
+[Vault operator charm](https://charmhub.io/vault-k8s) to securely store
+credentials that can be accessed by workflows. This is the recommended way of
+storing workflow-related credentials in production environments. To enable this,
+run the following commands:
+
+```bash
+juju deploy vault-k8s --channel edge
```
kian99, line 156: Is this the right channel? I see on [charmhub](https://charmhub.io/vault-k8s) the following,
Screenshot from 2024-06-13 08-43-14

README.md
```diff
@@ -137,14 +137,38 @@ juju relate temporal-worker-k8s admin/cos.loki
...
+
+# After following Vault doc instructions to unseal Vault
+juju relate temporal-worker-k8s vault-k8s
+```
+
+For a reference on how to access credentials from Vault through the workflow
```
kian99, line 162: Should there be mention here that you need to unseal the Vault, basically a link to the [tutorial](https://charmhub.io/vault-k8s/docs/h-getting-started?channel=1.15/edge).

kian99 (approved):

src/relations/vault.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        try:
+            secret = self.charm.model.get_secret(label=VAULT_NONCE_SECRET_LABEL)
+            nonce = secret.get_content(refresh=True)["nonce"]
+            return nonce
+        except ModelError as e:
+            logger.debug(f"Secret {VAULT_NONCE_SECRET_LABEL} not found: {e}")
+            raise ModelError from e
+
+    def _get_vault_config(self):
```
AmberCharitos, line 82: Does this need to be done by all units or could the leader do this and update state so that the peer relation changes and the other units update?

src/relations/vault.py
```diff
@@ -0,0 +1,112 @@
...
+        """
+        relation = self.charm.model.get_relation("vault")
+        if relation is None:
+            logger.debug("No vault relation found")
+            return None
+        vault_url = self.charm.vault.get_vault_url(relation)
+        ca_certificate = self.charm.vault.get_ca_certificate(relation)
+        mount = self.charm.vault.get_mount(relation)
+        unit_credentials = self.charm.vault.get_unit_credentials(relation)
+        if not unit_credentials:
```
AmberCharitos, line 96: If there is a vault relation but no unit credentials do we need to put this in a blocked status? Or happy for it to default in this case to not using Vault?

tests/unit/test_charm.py
```diff
@@ -208,6 +137,82 @@ def test_update_status_down(self, _process_wheel_file, _setup_container, mock_ex
...
+                    "exec": {"command": "python check_status.py"},
+                }
+            },
+        }
+
+        got_plan = harness.get_container_pebble_plan("temporal-worker").to_dict()
+        self.assertEqual(got_plan, want_plan)
+
+
+def add_vault_relation(test, harness):
```
AmberCharitos, line 182: It might be important to also validate that the vault relation gone away does indeed remove the vault credentials.

AmberCharitos (approved):
LGTM - nice work! I am sure this will be a useful reference for future charm vault integrations. A couple of suggestions/questions

requirements.txt
kelkawi-a: This is actually updating the version and pinning it to `2.14.0` from `2.2.0`.

resource_sample/dist/python_samples-1.1.0-py3-none-any.whl
kelkawi-a: Good point, updated.

README.md
kelkawi-a: Good catch, updated to `1.15/edge`.

src/relations/vault.py
kelkawi-a: The Vault relation does not offer a `relation_changed` hook of some sort so i'm not sure if the leader would update if anything related to the vault changes, even if this is included on the `on_ready` hook. I opted to do it the same way it is done in the JIMM charm just to be sure.

src/relations/vault.py
kelkawi-a: Good point :+1: Added logic to send it to a blocked state instead.

PR #19, opened 2024-06-21
Refactor charm to use ROCK image

kelkawi-a:
This PR introduces a major refactoring of the Temporal worker charm and the way it handles resources. Prior to this, the worker charm had a base Python image, and users would define their own workflows, build them as a wheel file using `poetry` and attach them to the charm. This approach had several problems, one of which is that it only works for Python. This PR does the following:

- Removes the wheel file resource, and replaces the `temporal-worker-image` resource with a ROCK image that is attached by the user at deployment time. Hence, the deployment would be done as `juju deploy temporal-worker-k8s --temporal-worker-image=<user_defined_ROCK>`. The charm now has two resources that the user would attach:
  - `temporal-worker-image`: This is a ROCK image that contains the user's worker and workflow definitions.
  - `env-file`: This is a `.env` file which contains environment variables to be injected into the workload container and can be used by the user's workflows.
- Refactors all tests to reflect the changes made.
- Adds a sample `resource_sample_py` with a `rockcraft.yaml` file which can be used to build the Python ROCK image. A future PR will include a `resource_sample_go` for workflows written using the Temporal Go SDK.
- Removes the `up` health check from the workload container as the function of the charm will be to inject the config variables into the workload container. If the workload container faces any issues when starting, the charm will go into an `error` state.

Note: The `test_upgrades.py` test file has been disabled for now as this PR introduces breaking changes, and therefore the recommendation is to redeploy the charm entirely rather than upgrading using `juju refresh`.

Note 2: Some CVEs have been ignored for now using `.trivyignore` with comments explaining why. This will be removed once the CVEs (which have already been resolved) are pushed to the latest builds and SDKs.


resource_sample_py/rockcraft.yaml
```diff
@@ -0,0 +1,75 @@
...
+name: temporal-worker
+summary: Temporal worker app
+description: OCI image for Temporal worker app
+version: "1.0"
+base: ubuntu@22.04
+build-base: ubuntu@22.04
+license: Apache-2.0
+
+services:
+  temporal-worker:
```
kian99, line 12: Will come back to review this properly but I had a fix an issue caused by JIMM's rock which might help save you some time, I recommend giving this PR a read https://github.com/canonical/jimm/pull/1247

kelkawi-a: There's a recommendation from the ROCKs team to include this so that the ROCK can be deployed without the charm. It should be fine here because the charm overrides this, but it's useful to have so users can verify the ROCK has been built correctly and functions as intended. (cc @AmberCharitos )

mertalpt (approved):
LGTM but I would have to see this in action before fully understanding it.

@kelkawi-a Is there any documentation we can link here (spec?) for the changes? Do we have a repository for our worker ROCKs now or do we need to create one? What are the steps to developing a new Temporal workflow now?

kelkawi-a:
> LGTM but I would have to see this in action before fully understanding it.
> 
> @kelkawi-a Is there any documentation we can link here (spec?) for the changes? Do we have a repository for our worker ROCKs now or do we need to create one? What are the steps to developing a new Temporal workflow now?

We will be using the same repository we currently have for our workflows, but will be updating them by adding the `rockcraft.yaml` file and adding the worker definition to the workflow itself, as it has been removed from the charm now. The steps haven't changed really, it's just a new way of attaching the resource to the charm which should become a more stable experience now.

The example under `resource_sample_py` shows how a workflow can be developed, built and deployed using the charm, and the `README.md` under that directory has some instructions to that extent.

gtato (approved):
looks good, juts a couple of questions, but nothing blocking

src/charm.py
```diff
@@ -396,19 +312,10 @@ def _update(self, event):  # noqa: C901
             "services": {
                 self.name: {
                     "summary": "temporal worker",
-                    "command": command,
+                    "command": "./app/scripts/start-worker.sh",
                     "startup": "enabled",
                     "override": "replace",
                     "environment": context,
-                    "on-check-failure": {"up": "ignore"},
```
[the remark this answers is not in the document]
kelkawi-a: This was to test that the worker process is up and running. Previously, the charm would go into an active state even though the worker failed to start, so this was added at the time. With the new setup, the charm goes into an error state if the worker process fails to start, so this is no longer needed.

PR #22, opened 2024-07-18
add secrets config and vault secret actions

kelkawi-a:
This PR introduces the following changes and additions:

- Prior to this, users would attach an `env-file` resource to specify a set of environment variables which should be injected into the workload container and used by the workflow definitions. This approach was problematic in that it handles secret credentials poorly. The `env-file` is now deprecated in favor of the `secrets` config described below.
- Adds the `secrets` config parameter, which can be used to instruct the charm to fetch secrets from three sources and inject them as environment variables into the workload container:
  - `env`: These are plaintext environment variables, which are usually not secret.
  - `juju` (Requires Juju 3.3+): These are variables stored as Juju secrets, where the user provides the ID of the secret and the key from which to fetch the value.
  - `vault`: These are variables fetched from Vault, provided the charm is related to a Vault charm.
- Removes the logic where the charm was injecting Vault connection parameters into the workload container, as the workflows ideally do not need to interact with Vault to fetch secrets. This work is now handled by the charm, the workload container should only expect specific environment variables to be present.
- Adds actions for adding and reading secrets from Vault.
- Updates docs and tests to reflect the changes.

Thanks to @gtato for the suggestions and useful design discussions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,19 +20,19 @@ connect to a deployed Temporal server.
 
 ### Deploying
 
-To deploy the Temporal Worker operator, you can start by creating a Temporal
-workflow, or use the one provided in
+To deploy the Temporal Worker operator charm, you can start by creating a
+Temporal workflow, or use the one provided in
 [`resource_sample_py`](./resource_sample_py/). Once done, the project can be
 built as a [ROCK](https://documentation.ubuntu.com/rockcraft/en/stable/) and
 pushed to the [local registry](https://microk8s.io/docs/registry-built-in) by
 running the following command inside the `resource_sample_py` directory:
 
 ```bash
-make build_rock
+make -C resource_sample_py build_rock
 ```
 
-The Temporal Worker operator can then be deployed and connected to a deployed
-Temporal server using the Juju command line as follows:
+The Temporal Worker operator charm can then be deployed and connected to a
+deployed Temporal server using the Juju command line as follows:
 
 ```bash
 juju deploy temporal-worker-k8s --resource temporal-worker-image=localhost:32000/temporal-worker-rock
@@ -60,20 +60,93 @@ Note: The only requirement for the ROCK is to have a `scripts/start-worker.sh`
 file, which will be used as the entry point for the charm to start the workload
 container.
 
-### Adding Environment Variables
+### Adding Secrets & Environment Variables
 
-The Temporal Worker operator can be used to inject environment variables that
-can be ingested by your workflows. This can be done using the Juju command line
-as follows:
+The Temporal Worker operator charm allows the user to configure multiple sources
+of environment variables and secrets to be injected into the workload container
+and consumed by the user's workflow definitions. These sources can be configured
+through the `secrets` config parameter of the charm. Below are the three sources
+of environment variables and secrets currently supported. A user may choose to
+use one, all or none of them. Once the `secrets.yaml` file is ready, it can be
+configured into the charm as follows:
 
 ```bash
-juju attach-resource temporal-worker-k8s env-file=path/to/.env
+juju config temporal-worker-k8s secrets=@/path/to/secrets.yaml
 ```
 
-#### **`.env`**
+These secrets can then be injested by the workflows by using the `os` package as
+follows:
 
+```python
+import os
+value1 = os.getenv("key1")
 ```
-VALUE=123
+
+#### Direct Environment Variables
+
+These are usually values that are not secret and can be stored as plaintext. An
+example is setting the application environment to `staging` or `production`.
+They can be set as follows:
+
+##### **`secrets.yaml`**
+
+```yaml
+secrets:
+  env:
+    - key1: value1
+    - key2: value2
+```
+
+#### Juju User Secrets (Requires Juju 3.3+)
+
+[Juju secrets](https://juju.is/docs/juju/manage-secrets) are values which can be
+stored in the model and accessed by the charm. To do so, you must first add the
+secret and grant the charm access to it:
+
+```bash
+juju add-secret my-secret key1=value1 key2=value2
+
+# Output: secret:<secret_id>
+
+juju grant-secret my-secret temporal-worker-k8s
+```
+
+The secrets can then be configured into the charm as follows:
+
+##### **`secrets.yaml`**
+
+```yaml
+secrets:
+  juju:
+    - secret-id: <secret_id>
+      key: key1
+    - secret-id: <secret_id>
+      key: key2
+```
+
+#### Vault
+
+The Vault section below outlines how the Temporal Worker operator charm can be
+related to the Vault operator charm for storing secrets securely. Once done, the
+charm can be configured to fetch secrets from Vault and inject them as variables
+into the workload container. The secrets can be configured into the charm as
+follows:
+
+##### **`secrets.yaml`**
+
+```yaml
+secrets:
+  vault:
+    - path: my-secrets
+      key: key1
+    - path: my-secrets
+      key: key2
+```
+
+These secrets can then be added to Vault by running the following charm action:
+
+```bash
+juju run temporal-worker-k8s/leader add-vault-secret path="my-secrets" key="key1" value="value1"
 ```
 
 ## Verifying
@@ -112,9 +185,9 @@ juju scale-application temporal-worker-k8s <num_of_replicas_required_replicas>
 
 ## Error Monitoring
 
-The Temporal Worker operator has a built-in Sentry interceptor which can be used
-to intercept and capture errors from the Temporal SDK. To enable it, run the
-following commands:
+The Temporal Worker operator charm has a built-in Sentry interceptor which can
+be used to intercept and capture errors from the Temporal SDK. To enable it, run
+the following commands:
 
 ```bash
 juju config temporal-worker-k8s sentry-dsn=<YOUR_SENTRY_DSN>
@@ -163,7 +236,7 @@ storing workflow-related credentials in production environments. To enable this,
 run the following commands:
 
 ```bash
-juju deploy vault-k8s --channel 1.15/edge
+juju deploy vault-k8s --channel 1.16/edge
 
 # After following Vault doc instructions to unseal Vault
 juju relate temporal-worker-k8s vault-k8s
--- a/README.md
+++ b/README.md
@@ -65,17 +65,17 @@ container.
 The Charmed Temporal Worker allows the user to configure multiple sources of
 environment variables and secrets to be injected into the workload container and
 consumed by the user's workflow definitions. These sources can be configured
-through the `secrets` config parameter of the charm. Below are the three sources
-of environment variables and secrets currently supported. A user may choose to
-use one, all or none of them. Once the `secrets.yaml` file is ready, it can be
-configured into the charm as follows:
+through the `environment` config parameter of the charm. Below are the three
+sources of environment variables and secrets currently supported. A user may
+choose to use one, all or none of them. Once the `environment.yaml` file is
+ready, it can be configured into the charm as follows:
 
 ```bash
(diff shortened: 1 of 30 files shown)


AmberCharitos (commented):
Great work! This is really nice, some non-blocking suggestions and one important question around the use of temporal actions to add and remove vault secrets.

src/relations/vault.py
```diff
@@ -105,11 +120,47 @@ def _get_vault_config(self):
...
+        """
+        ca_certificate_path = self.get_ca_cert_location_in_charm()
+        vault_config = self.get_vault_config()
+        return VaultClient(
+            address=vault_config["vault_address"],
+            cert_path=f"{ca_certificate_path}/{VAULT_CA_CERT_FILENAME}",
+            role_id=vault_config["vault_role_id"],
+            role_secret_id=vault_config["vault_role_secret_id"],
+            mount_point=vault_config["vault_mount"],
+        )
```
onurmus, line 150: [Suggestion] Here you can use spread annotation like the below:

```
        return {
            "address": vault_url,
            "role_id": role_id,
            "role_secret_id": role_secret_id,
            "mount": mount,
        }

    def get_vault_client(self):
        """Initialize Vault client.

        Returns:
            Vault client.
        """
        ca_certificate_path = self.get_ca_cert_location_in_charm()
        vault_config = self.get_vault_config()
        return VaultClient(
            **{**vault_config, 'cert_path': f"{ca_certificate_path}/{VAULT_CA_CERT_FILENAME}"}
        )
```

`vault_ca_certificate_bytes` and `vault_cert_path` are not used anywhere so you can remove them from return dictionary.

src/relations/vault.py
```diff
@@ -105,11 +120,47 @@ def _get_vault_config(self):
...
+        Returns:
+            Path: The CA certificate location
+        """
+        storage = self.charm.model.storages
+        if "certs" not in storage:
+            return None
+        if len(storage["certs"]) == 0:
+            return None
+        cert_storage = storage["certs"][0]
+        return cert_storage.location
```
onurmus, line 166: [Suggestion] Could be a bit shorter:

```
if not storage.get("certs"):
    return None
return storage["certs"][0].location
```

src/vault/actions.py
```diff
@@ -0,0 +1,97 @@
...
+            event.fail("No vault relation found")
+
+        path, key, value = (event.params.get(param) for param in ["path", "key", "value"])
+        if not all([path, key, value]):
+            event.fail("`path`, `key` and `value` are required parameters")
+
+        try:
+            vault_client = self.charm.vault_relation.get_vault_client()
+        except Exception:
+            event.fail("unable to initialize vault client. remove relation and retry.")
```
onurmus, line 52: [Typo] uppercase "unable"

src/vault/actions.py
```diff
@@ -0,0 +1,97 @@
...
+
+        try:
+            vault_client.write_secret(path=path, key=key, value=value)
+            self.charm._update(event)
+        except ValueError as e:
+            logger.error("unable to create secret in vault: %s", str(e))
```
onurmus, line 58: [Typo] uppercase "unable"

src/vault/actions.py
```diff
@@ -0,0 +1,97 @@
...
+        Args:
+            event:The event triggered by the restart action
+        """
+        container = self.charm.unit.get_container(self.charm.name)
+        if not container.can_connect():
+            event.fail("Failed to connect to the container")
+            return
+
+        if not self.charm.model.relations["vault"]:
+            event.fail("No vault relation found")
```
onurmus, line 77: [Suggestion] Had the same block for `_on_add_vault_secret` method. You may create a method called `validate_relation`.

tests/integration/helpers.py
```diff
@@ -21,11 +22,95 @@
...
+            await ops_test.model.applications[APP_NAME]
+            .units[0]
+            .run_action("add-vault-secret", path=path, key=key, value=value)
+        )
+        result = (await action.wait()).results
+        logger.info("add vault secret result: %s", result)
+        if "result" in result and result["result"] == "secret successfully created":
+            secret_added = True
+            break
+        time.sleep(2)
```
onurmus, line 110: Why do we need this? If the computer is slow, may we need 10 seconds?

onurmus (approved):
I left some small comments but none of them is a blocker. Great work!

src/vault/actions.py
```diff
@@ -0,0 +1,97 @@
...
+            charm: The charm to attach the hooks to.
+        """
+        super().__init__(charm, "vault-actions")
+        self.charm = charm
+
+        charm.framework.observe(charm.on.add_vault_secret_action, self._on_add_vault_secret)
+        charm.framework.observe(charm.on.get_vault_secret_action, self._on_get_vault_secret)
+
+    @log_event_handler(logger)
+    def _on_add_vault_secret(self, event):
```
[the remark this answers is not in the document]
kelkawi-a: This is a case of separation of access v practicality. We had a similar discussion for the Temporal charm when adding the OpenFGA relation in [this PR](https://github.com/canonical/temporal-k8s-operator/pull/23), where we added utility actions for interacting with the OpenFGA store through the Temporal charm.

There is a way of doing it directly through the vault snap, but not through actions on the Vault charm. I opted to go down a similar path as the Temporal charm and add some utility methods that allow users to easily add secrets to Vault through the Temporal worker charm.

tests/integration/helpers.py
kelkawi-a: This is just to leave a small duration of time between each action attempt. The action generally works from the first attempt, but this is only here is a failsafe.

src/secret_processors.py
```diff
@@ -0,0 +1,169 @@
...
+        raise ValueError(
+            "Invalid secrets structure: 'juju' should be a list of dictionaries with 'key' and 'secret-id'"
+        )
+
+    # Validate vault key
+    vault = secrets_key.get("vault", [])
+    if not isinstance(vault, list) or not all(
+        isinstance(item, dict) and "path" in item and "key" in item and len(item) == 2 for item in vault
+    ):
+        raise ValueError("Invalid secrets structure: 'vault' should be a list of dictionaries with 'path' and 'key'")
```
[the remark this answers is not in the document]
kelkawi-a: Good point, I'll leave that as a TODO for a separate PR

README.md
```diff
@@ -60,20 +60,93 @@ Note: The only requirement for the ROCK is to have a `scripts/start-worker.sh`
...
 ```
-VALUE=123
+
+#### Direct Environment Variables
+
+These are usually values that are not secret and can be stored as plaintext. An
+example is setting the application environment to `staging` or `production`.
+They can be set as follows:
+
+##### **`secrets.yaml`**
```
gtato, line 91: what do you think about naming this `environment.yaml` and setting `env` as root (since all secrets are env but not vice versa)
a low impact change would be:
```
env:
    plain (or default):
        - key1: value1
    juju:
       ....
    vault:
        ...    
```
we could opt for a k8s-ish syntax too (more impact):
```
env:
    - name: key1
      value: value1
    - name: key2
      valueFromSecret:
          type: juju
          id: secret-id
          key: key2
    - name: key3
      valueFromSecret:
          type: vault
          id: secret-path
          key: key3    
```
or a more compact form:
```
env:
    - name: key1
      value: value1
    - name: key2
      valueFromSecret: juju@secret-id:key2
    - name: key3
      valueFromSecret: vault@secret-path:key3    
```

README.md
```diff
@@ -60,20 +60,93 @@ Note: The only requirement for the ROCK is to have a `scripts/start-worker.sh`
...
+    - path: my-secrets
+      key: key1
+    - path: my-secrets
+      key: key2
+```
+
+These secrets can then be added to Vault by running the following charm action:
+
+```bash
+juju run temporal-worker-k8s/leader add-vault-secret path="my-secrets" key="key1" value="value1"
```
gtato, line 149: can you add multiple keys for the same path here?

README.md
```diff
@@ -60,20 +60,93 @@ Note: The only requirement for the ROCK is to have a `scripts/start-worker.sh`
...
+
+The secrets can then be configured into the charm as follows:
+
+##### **`secrets.yaml`**
+
+```yaml
+secrets:
+  juju:
+    - secret-id: <secret_id>
+      key: key1
```
gtato, line 122: does this mean that the name of the environment variable needs to match the name of the secret in juju or vault? There might be cases where that name is not known before hand.

src/secret_processors.py
```diff
@@ -0,0 +1,171 @@
...
+            try:
+                secret = vault_client.read_secret(path=path, key=key)
+            except Exception as e:
+                raise ValueError(f"Unable to read vault secret `{key}` at path `{path}`: {e}") from e
+            charm_env.update({key: secret})
+
+    return charm_env
+
+
+def parse_secrets(yaml_string):
```
gtato, line 106: maybe we should use pydantic here?

src/vault/actions.py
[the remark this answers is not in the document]
gtato: if this is required at the moment to make the charm functional, ok, but i agree with Amber that this defies a bit the point of Valut. If the same party can read and write secrets in/from Vault, there is practically no difference between using vault or an environment variable. If we have a centralized vault, we'd assume that different people add secrets potentially via the UI.

gtato (approved):
Left some comments, nothing blocking.

README.md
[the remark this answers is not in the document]
kelkawi-a: Yes

src/secret_processors.py
kelkawi-a: I've added a TODO for converting the charm to use structured config instead

README.md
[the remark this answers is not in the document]
kelkawi-a: Not sure I understand but yes the key in the secret must match the key that the user specified in their code. What would be an alternative?

README.md
```diff
@@ -60,20 +60,93 @@ Note: The only requirement for the ROCK is to have a `scripts/start-worker.sh`
...
+#### Direct Environment Variables
+
+These are usually values that are not secret and can be stored as plaintext. An
+example is setting the application environment to `staging` or `production`.
+They can be set as follows:
+
+##### **`secrets.yaml`**
+
+```yaml
+secrets:
```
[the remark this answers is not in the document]
kelkawi-a: Agreed. I went with the low impact change, as I'd like to avoid having the config look k8s-ish, and the compact form might have readability issues.

README.md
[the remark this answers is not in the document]
gtato: this is a bit related to the convo Amber touched on. People who set secrets should in general be different from the ones who write the code. So in the code you know you need some environment variable, but you should be able to call it whatever you want (say `MY_ENV`). Similarly, the person who writes the secret, can put a whatever key (say `my_secret`). Only at the moment of the deployment, you should say: load `my_secret` value into `MY_ENV`. We should not assume that the programmer knows that the key is `my_secret` and do `os.getenv('my_secret')` from the beginning.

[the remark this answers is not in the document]
kelkawi-a: Good point. So for juju and Vault, I guess we will need to add an optional key of what the env variable name should be? Something like:

```yaml
juju:
   - secret-id: <secret_id>
     key: key
     dest-key: MY_ENV
```

Not sure what this optional key should be called, any suggestions?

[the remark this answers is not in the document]
gtato: that is fine, but then i think you would need to keep the *env* consistent:
```
env:
   - name: MY_ENV1
     value: bla bla
juju:
   - name: MY_ENV2
     secret-id: secret-id
     key: key2
vault:
   - name: MY_ENV3
     secret-path: secret-path
     key: key3
```

PR #26, opened 2024-09-10
Fixed cant find rust compiler error

onurmus:
* I got `error: can't find Rust compiler` error while trying to pack the charm in my multipass VM. After a bit of investigation, I see that it's related to `cosl` package. To solve the problem I added rust toolchain. Here is the [reference issue](https://github.com/canonical/bundle-kubeflow/issues/989). It seems that we are not alone with this error 😄 
* Fixed ops-scenario version for tests because with the latest version of ops-scenario(7.x) we are getting `cannot import name 'Event' from 'scenario' ` [error](https://github.com/canonical/pytest-interface-tester/issues/23). Also, we were getting the same error when building the charm. 
* Ignored `CVE-2024-45311`(September 2) because it relates to `temporalio` SDK. 
* Ignored `CVE-2024-34156` which is a new issue relating to Go and the Foundations team [working on it](https://warthogs.atlassian.net/browse/FR-8847).
* Changed Contributing.MD so that we use microk8s strict version which is required by juju 3.5. 
* Changed Contributing.MD so that docker commands run without permission issues.
* Changed Makefile to simplify rock building and push to local registry.


gtato (approved):
lg (might need to update libs)

requirements.txt
kelkawi-a, line 1: EOF on new line.

kelkawi-a (approved):
LGTM, thanks for working on this!

kelkawi-a:
> lg (might need to update libs)

It looks like the Vault library is the one needing updates. Might be better fit for a separate PR to avoid increasing the scope.

PR #28, opened 2024-09-15
Fixed bug for  overriding rock environment variables

onurmus:
We had a bug about during updating environment variables of temporal worker layer.   

## Case
 You try to set `TWC_AUTH_PROVIDER='candid'` for your worker rock image using `juju config temporal-worker-k8s environment=@./config.yaml` command.

## Bug
In `_update` method in `charm.py` file, we first get environment variables in [line 259 ](https://github.com/canonical/temporal-worker-k8s-operator/blob/e3767a360c1f6fa640f8e34d9f411917ca0ece3f/src/charm.py#L259). Then, we override the environment variables with [charm's config parameters](https://github.com/canonical/temporal-worker-k8s-operator/blob/e3767a360c1f6fa640f8e34d9f411917ca0ece3f/src/charm.py#L280). 

## Reason
Probably we haven't seen this bug because we provide same config to charm and rock images. But in case you want to have no authentication(meaning does not provide `TWC_AUTH_PROVIDER`), the charm overrides it with its default config.

## Solution
The solution is that first read charm configurations and then override it with rocks configuration.

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -254,6 +254,9 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             return
 
         context = {}
+        context.update(
+            {convert_env_var(key): value for key, value in self.config.items() if key not in ["environment"]}
+        )
         try:
             self._validate(event)
             environment_config = self.config.get("environment")
@@ -277,9 +280,6 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             if value:
                 context.update({key: value})
 
-        context.update(
-            {convert_env_var(key): value for key, value in self.config.items() if key not in ["environment"]}
-        )
         context.update({"TWC_PROMETHEUS_PORT": PROMETHEUS_PORT})
 
         pebble_layer = {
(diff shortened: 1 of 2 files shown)


kelkawi-a:
I'm not sure I fully understood the problem. The charm is meant to provide the rock image with configs and environment variables. If you want to have no authentication, it would be sufficient to set the charm config `auth-provider` to `""`. Am I missing something?

gtato (commented):
I'm not sure I understand the use case. From what I understand the environment variables are something used by the workflow itself, while the config are for worker and are common to all workflows... in theory there should not be any overlap between the two.

src/charm.py
```diff
@@ -254,6 +254,9 @@ def _update(self, event):  # noqa: C901
             return
 
         context = {}
+        context.update(
+            {convert_env_var(key): value for key, value in self.config.items() if key not in ["environment"]}
```
mertalpt, line 258: question: Why not `key != "environment"`? Did you meant to do something else here?

E: I see that this line just changed places. Still, if you (or the author) can tell why it is written this way that would be nice.

mertalpt (approved):

src/charm.py
kelkawi-a: All config variables on the charm are being converted into environment variables to be injected into the workload container, with the exception of the `environment` config which is a yaml-based config for configuring secrets introduced in #22 .

onurmus:
After discussing with @kelkawi-a and @gtato, I realized that "It's not a bug, it's a feature" 😄  We expect to have the charm to maintain environment variables starting with TWC. That also means that each worker charm will need to have one rock image attached. My assumption was that we deploy different workers to the same charm. Sorry for the confusion.

PR #29, opened 2024-09-16
Enable providing list and dict variables in env variables config

onurmus:
### Use Case
In the current implementation, we can just provide environment variables which are text, numeric, or boolean values. However, I need to provide a transformation config which is a nested configuration. In this case, the charm tries to provide a JSON object as an environment variable. 

```
env:
  - name: TWC_HOST
    value: 192.168.67.5:7233
  - name: TWC_QUEUE
    value: test
  - name: TWC_NAMESPACE
    value: default
  - name: WF_TRANSFORMATION
    value:
      - connection_id: my_connection_id
        unnesting:
          tables:
            issues: ["test1", "test2"]
            users: [test3]
        redaction:
```

### Solution
Use `json.dumps()` to convert `dict` or `list`  to string.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -170,6 +170,14 @@ options:
             env:
               - name: key1
                 value: value1
+              - name: nested_example_key
+                value: 
+                  - connection_id: a_connection_id
+                    unnesting:
+                      tables:
+                        issues: ["test1", "test2"]
+                        users: [test3]
+                    redaction:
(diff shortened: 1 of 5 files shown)


gtato (approved):
lg

kelkawi-a (approved):
LGTM, thanks for working on this!

PR #30, opened 2024-10-02
Add initialization through env variables functionality

kelkawi-a:
This PR does the following:
- Injects `TEMPORAL_*` environment variables from config, which can now be consumed by the [Temporal client library](https://github.com/canonical/temporal-lib-py/tree/main/temporallib) after merging [this PR](https://github.com/canonical/temporal-lib-py/pull/15). This means the Temporal worker can be initialized through the charm without having to specify the parameters. An example can be found below:

- Returns a check that was removed earlier, which restricts users from setting environment variables with the prefix `TEMPORAL_*` or `TWC_*` as they are reserved by the charm.


onurmus (approved):
LG!

mertalpt (approved):

PR #32, opened 2024-11-24
Add PostgreSQL DB relation

kelkawi-a:
This PR adds a relation between the Temporal worker charm and the PostgreSQL charms through the `postgresql_client` interface. This can be used to allow users to create and interact with a database from within their workflow code.

It also updates the `vault` and `loki` libraries.

### DIFF
--- a/resource_sample_py/resource_sample/activities/db_activity.py
+++ b/resource_sample_py/resource_sample/activities/db_activity.py
@@ -6,6 +6,7 @@ import os
 import psycopg2
 from psycopg2 import sql
 from common.messages import ComposeGreetingInput
+from psycopg2.extras import RealDictCursor
 
 
 @activity.defn(name="database_test")
@@ -18,7 +19,7 @@ async def database_test(arg: ComposeGreetingInput) -> str:
         "port": os.getenv("TEMPORAL_DB_PORT")
     }
     conn = psycopg2.connect(**db_config)
-    cursor = conn.cursor()
+    cursor = conn.cursor(cursor_factory=RealDictCursor)
     
     # Create test table
     table_name = "test_table"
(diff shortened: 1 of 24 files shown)


resource_sample_py/resource_sample/activities/db_activity.py
```diff
@@ -0,0 +1,58 @@
...
+    
+    # Read the record back
+    select_query = sql.SQL("SELECT * FROM {table} WHERE id = %s").format(
+        table=sql.Identifier(table_name)
+    )
+    cursor.execute(select_query, (inserted_id,))
+    record = cursor.fetchone()
+
+    # Close the cursor and connection
+    if cursor:
```
mertalpt, line 53: suggestion: Use `with` in the sample for safer code. [Ref](https://www.psycopg.org/docs/connection.html).

src/relations/postgresql.py
```diff
@@ -0,0 +1,119 @@
...
+            return
+
+        if not self.charm._state.is_ready():
+            event.defer()
+            return
+
+        self.charm.unit.status = WaitingStatus(f"handling {event.relation.name} change")
+
+        self.update_db_relation_data_in_state(event)
+        self.charm._update(event)
```
mertalpt, line 49: We talked about (emitting custom events instead of calling) but I am not going to push it if you want to keep it like this.

mertalpt (approved):
Note the Trivy notices for `.trivyignore`.

resource_sample_py/resource_sample/activities/db_activity.py
```diff
@@ -0,0 +1,58 @@
...
+import os
+import psycopg2
+from psycopg2 import sql
+from common.messages import ComposeGreetingInput
+from psycopg2.extras import RealDictCursor
+
+
+@activity.defn(name="database_test")
+async def database_test(arg: ComposeGreetingInput) -> str:
+    db_config = {
```
gtato, line 14: i guess we can pydantic this, and provide a singleton of connection from a centralized location (maybe shared libs...).

gtato (approved):
looks good, just a non-blocking comment

PR #37, opened 2025-02-13
allow the config options of the temporal connection to be overridden by juju and vault secrets

### LINKED ISSUE #36: Cannot override temporal config files using environmental variables.
According to https://charmhub.io/temporal-worker-k8s/configurations#environment, the environmental variables are read in order of vault -> juju -> environmental variables.

However, this means that config options in the charm i.e `TEMPORAL_ENCRYPTION_KEY` cannot be overridden using the higher priority variable storage. i.e the following config will not respect the `TEMPORAL_ENCRYPTION_KEY` from the juju secret, rather it will be reset to an empty `TEMPORAL_ENCRYPTION_KEY`

```yaml
juju:
    - secret-id: {secret_id}
      name: TEMPORAL_ENCRYPTION_KEY
      key: encryption-key
vault:
    - path: secrets
      name: access_token
      key: token
```

brotich:
With the current setup of the charm the config options i.e TEMPORAL_ENCRYPTION_KEY cannot be overridden by the value store in the vault or juju secrets. it always defaults to the value set on the charm even after is has been defined in a juju secret or a vault.

This change adds a check for the already defined environmental variables in the context to avoid the `convert_env_var` setting the `TWC_` and `TEMPORAL_` prefix variable from being set to empty. see the test for the example.

fixes #36

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -303,21 +303,18 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             if value:
                 context.update({key: value})
 
-        context.update(
-            {
-                convert_env_var(key, prefix="TWC_"): value
-                for key, value in self.config.items()
-                if key not in ["environment"]
-            }
-        )
+        for key, value in self.config.items():
+            if key in ["environment"]:
+                continue
 
-        context.update(
-            {
-                convert_env_var(key, prefix="TEMPORAL_"): value
-                for key, value in self.config.items()
-                if key not in ["environment"]
-            }
-        )
+            # Ignore environmental variables that have been overridden
+            # by vault or juju config options.
+            twc_env_var = convert_env_var(key, prefix="TWC_")
+            if not is_env_var_overridden(context, twc_env_var):
+                context.update({twc_env_var: value})
+            temporal_env_var = convert_env_var(key, prefix="TEMPORAL_")
+            if not is_env_var_overridden(context, temporal_env_var):
+                context.update({temporal_env_var: value})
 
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -303,21 +303,18 @@ def _update(self, event):  # noqa: C901
...
-        )
-
-        context.update(
-            {
-                convert_env_var(key, prefix="TEMPORAL_"): value
-                for key, value in self.config.items()
-                if key not in ["environment"]
-            }
-        )
+        for key, value in self.config.items():
```
mertalpt, line 306: q: These sections look like they do the same thing. What is the difference?

src/charm.py
```diff
@@ -303,21 +303,18 @@ def _update(self, event):  # noqa: C901
...
+            if key in ["environment"]:
```
frankban, line 307: if key == "environment"

src/charm.py
```diff
@@ -303,21 +303,18 @@ def _update(self, event):  # noqa: C901
...
+                continue
+
+            # Ignore environmental variables that have been overridden
```
frankban, line 310: # Ignore empty environmental variables

src/charm.py
```diff
@@ -303,21 +303,18 @@ def _update(self, event):  # noqa: C901
...
+            # by vault or juju config options.
+            twc_env_var = convert_env_var(key, prefix="TWC_")
+            if not is_env_var_overridden(context, twc_env_var):
```
frankban, line 313: if not context.get(twc_env_var):

frankban (approved):
Looks good. I think this is only introducing the notion that if a variable is empty then it is considered unset, and taken, if possible, at a higher level.

brotich:
closing as this might not be the way forward to implement this.

PR #38, opened 2025-03-19
Improve juju secrets handling

kelkawi-a:
This PR improves the way juju secrets are handled to make it less cumbersome to configure the worker charm. Previously, operators would have to configure the charm to read each key as follows:

```yaml
env:
    juju:
        - secret-id: {secret_id}
          name: ACCESS_TOKEN
          key: access-token
        - secret-id: {secret_id}
          name: CLIENT_SECRET
          key: client-secret
```

However, this approach can quickly become unscalable. This PR now allows operators to simply specify the secret ID as follows:

```yaml
env:
    juju:
        - secret-id: {secret_id}
```

The charm will then read all the key/value pairs in this secret, convert them into environment variables and inject them into the workload container.

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,2 +1,2 @@
-# TODO (kelkawi-a): remove these once pebble CVE resolution propagates down the pipeline
-CVE-2024-45338
+# TODO (kelkawi-a): remove these once temporal propagates the fix to their SDK
+CVE-2025-29787
--- a/README.md
+++ b/README.md
@@ -107,7 +107,7 @@ secret and grant the charm access to it:
 ```bash
 juju add-secret my-secret key1=value1 key2=value2
 
-# Output: secret:<secret_id>
+# Output: secret:<secret_id1>
 
 juju grant-secret my-secret temporal-worker-k8s
 ```
@@ -118,12 +118,13 @@ The environment variables can then be configured into the charm as follows:
 
 ```yaml
 juju:
-  - secret-id: <secret_id>
+  - secret-id: <secret_id1>
     name: env_var1
     key: key1
-  - secret-id: <secret_id>
+  - secret-id: <secret_id1>
     name: env_var2
     key: key2
+  - secret-id: <secret_id2> # reads all keys from this secret
 ```
 
 #### Vault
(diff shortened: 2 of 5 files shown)


.trivyignore
```diff
@@ -1,2 +1,2 @@
-# TODO (kelkawi-a): remove these once pebble CVE resolution propagates down the pipeline
-CVE-2024-45338
+# TODO (kelkawi-a): remove these once temporal propagates the fix to their SDK
```
mertalpt, line 1: suggestion: Fulfill the TODO.

config.yaml
```diff
@@ -183,12 +183,13 @@ options:
                       users: [test3]
                   redaction:
           juju:
-            - secret-id: <secret_id>
+            - secret-id: <secret_id1>
               name: sensitive1
```
mertalpt, line 187: q: Are these not supposed to be removed?

README.md
```diff
@@ -118,14 +120,20 @@ The environment variables can then be configured into the charm as follows:
 
 ```yaml
 juju:
-  - secret-id: <secret_id>
+  - secret-id: <secret_id1>
     name: env_var1
```
mertalpt, line 124: q: Are these not supposed to be removed?

tests/unit/test_charm.py
```diff
@@ -172,7 +171,7 @@ def test_invalid_environment_config(self):
         self.assertEqual(
             harness.model.unit.status,
             BlockedStatus(
-                "Invalid environment structure: 'juju' should be a list of dictionaries with 'secret-id', 'name', and 'key'"
+                "Invalid environment structure: each item in 'juju' must either contain only 'secret-id' or all of 'secret-id', 'name', and 'key'"
```
mertalpt, line 174: thought: This is a bit verbose. How good of an UX would it be to keep the message to `Invalid environment structure` and leave the rest to the logs?

mertalpt (commented):
I will approve after documentation comments are addressed.

.trivyignore
kelkawi-a: This is a new TODO, the old one was resolved.

RenanGreca (approved):

mertalpt (approved):

PR #40, opened 2025-04-10
Move auth configs to Juju user secrets

kelkawi-a:
The Temporal worker charm was first developed prior to the introduction of [Juju user secrets](https://documentation.ubuntu.com/juju/latest/howto/manage-secrets/index.html), hence why sensitive credentials are stored in charm configuration options. This PR moves sensitive authentication and encryption values into a Juju user secret, with the user only having to configure the charm with this secret's ID.


config.yaml
```diff
@@ -59,14 +59,20 @@ options:
...
 
+  auth-secret-id:
+    description: |
+      Juju secret ID containing authentication and encryption key parameters. This takes precedence over
+      the auth-related configuration options, which will be deprecated in the future.
+    type: string
+
   encryption-key:
-    description: Base64-encoded key used for data encryption.
+    description: (To be deprecated in favor of 'auth-secret-id') Base64-encoded key used for data encryption.
```
RenanGreca, line 69: suggestion: you can just say "deprecated" instead of "to be deprecated", since these fields are now planned for obsolescence. We do not want to encourage further use.

src/literals.py
```diff
@@ -23,3 +23,21 @@
 ]
 SUPPORTED_AUTH_PROVIDERS = ["candid", "google"]
 PROMETHEUS_PORT = 9000
+AUTH_SECRET_PARAMETERS = [
```
RenanGreca, line 26: q: shouldn't the Sentry DSN also be here? Maybe also the other Sentry parameters?

src/literals.py
```diff
@@ -23,3 +23,21 @@
...
+    "encryption-key",
+    "auth-provider",
```
RenanGreca, line 28: thought: `auth-provider`, `oidc-auth-type` and a couple of others are not secrets. I think it makes sense to have them grouped together.

RenanGreca (commented):
Thanks for the quick work! Left a couple of notes/questions.

src/literals.py
kelkawi-a: I was under the impression that Sentry DSN should be kept private, but other teams are of a different opinion. I opted to leave the Sentry parameters as is and leave it up to the discretion of the Workflows team.

src/literals.py
kelkawi-a: I chose to lump them together in the auth-secret out of convenience, having all auth-related params in one place.

mertalpt (approved):
Do we have any users that we have to support and who use Juju<3.3? If not, we can just remove the configuration method entirely instead of deprecating them. People can not update revisions if they cannot update their deployments yet.

@kelkawi-a What do you think?

RenanGreca (approved):

kelkawi-a:
> Do we have any users that we have to support and who use Juju<3.3? If not, we can just remove the configuration method entirely instead of deprecating them. People can not update revisions if they cannot update their deployments yet.
> 
> @kelkawi-a What do you think?

I thought it would be better not to break deployments if someone accidentally upgraded their deployment. They can be kept for a bit more before being completely removed.

PR #41, opened 2025-05-21
Add support for workload metrics

yakupt-canonical:
Currently, Temporal worker charm only supports metrics endpoint for Temporal SDK metrics, it does not support application defined metrics endpoint.

To enable application metrics, this PR adds a relation for workload metrics endpoint.

### DIFF
--- a/config.yaml
+++ b/config.yaml
@@ -10,7 +10,7 @@ options:
   # An example config option to customise the log level of the workload
   log-level:
     description: |
-      Configures the log level of gunicorn. 
+      Configures the log level of gunicorn.
 
       Acceptable values are: "info", "debug", "warning", "error" and "critical"
     default: "info"
@@ -53,9 +53,9 @@ options:
 
   sentry-sample-rate:
     description: |
-      A value between 0 (0% of errors) and 1 (100% of errors) to indicate the proportion of errors 
+      A value between 0 (0% of errors) and 1 (100% of errors) to indicate the proportion of errors
       to be captured by Sentry.
-    
+
     default: 1.0
     type: float
 
@@ -73,7 +73,7 @@ options:
   auth-provider:
     description: |
       (Deprecated in favor of 'auth-secret-id') Authentication provider for user authentication.
-    
+
       Acceptable values are "candid" and "google".
     default: ""
     type: string
@@ -158,30 +158,34 @@ options:
     default: ""
     type: string
 
+  workload-prometheus-port:
+    description: Port of the workload defined metrics endpoint.
+    type: int
+
   environment:
     description: |
-      This configuration is used to manage and retrieve sensitive information required 
-      by the application from different sources. The `environment` configuration supports 
+      This configuration is used to manage and retrieve sensitive information required
+      by the application from different sources. The `environment` configuration supports
       the following sources:
-      
+
       - **Environment Variables**: Plaintext environment variables.
       - **Juju**: Secrets can be managed and retrieved using Juju's secret storage capabilities.
       - **Vault**: Secrets can be securely stored and accessed from a HashiCorp Vault instance.
-      
-      The application will prioritize these sources in the following order: Vault, Juju, 
-      and then environment variables. If a variable is not found in the higher priority 
-      sources, it will fallback to the next available source. This ensures that the 
-      application can function correctly in various deployment scenarios while maintaining 
+
+      The application will prioritize these sources in the following order: Vault, Juju,
+      and then environment variables. If a variable is not found in the higher priority
+      sources, it will fallback to the next available source. This ensures that the
+      application can function correctly in various deployment scenarios while maintaining
       security and flexibility.
 
       Sample structure:
-    
+
         ```yaml
           env:
             - name: key1
               value: value1
             - name: nested_example_key
-              value: 
+              value:
                 - connection_id: a_connection_id
                   unnesting:
                     tables:
--- a/src/charm.py
+++ b/src/charm.py
@@ -81,6 +81,15 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             refresh_event=self.on.config_changed,
         )
 
+        if self.config.get("workload-prometheus-port"):
+            WORKLOAD_PROMETHEUS_PORT = self.config.get("workload-prometheus-port")
+            self._app_prometheus_scraping = MetricsEndpointProvider(
+                self,
+                relation_name="workload-metrics-endpoint",
+                jobs=[{"static_configs": [{"targets": [f"*:{WORKLOAD_PROMETHEUS_PORT}"]}]}],
+                refresh_event=self.on.config_changed,
+            )
+
         # Loki
         self._log_forwarder = LogForwarder(self, relation_name="logging")
 
--- a/metadata.yaml
+++ b/metadata.yaml
@@ -9,11 +9,11 @@ display-name: Temporal Worker
 summary: Temporal Worker operator
 description: |
   Temporal is a developer-first, open source platform that ensures
-  the successful execution of services and applications (using workflows). 
-  
-  This charm provides the Temporal worker which can connect to the Temporal 
+  the successful execution of services and applications (using workflows).
+
+  This charm provides the Temporal worker which can connect to the Temporal
   server via https to execute workflows.
-maintainers: 
+maintainers:
   - Commercial Systems <jaas-crew@lists.canonical.com>
 source: https://github.com/canonical/temporal-worker-k8s-operator
 docs: https://github.com/canonical/temporal-worker-k8s-operator
@@ -51,6 +51,8 @@ provides:
     interface: prometheus_scrape
   grafana-dashboard:
     interface: grafana_dashboard
+  workload-metrics-endpoint:
+    interface: prometheus_scrape
 
 requires:
   logging:


config.yaml
```diff
@@ -158,30 +158,34 @@ options:
     default: ""
     type: string
 
+  workload-prometheus-port:
```
gaborbk, line 161: Should we mark this value as not required?

gaborbk (commented):
Looks good, did you have the chance to try this locally? Just to make sure the charm starts and works as expected.

kelkawi-a (commented):
I believe this requires more thought on the requirements part. What is the end goal here? An operator should be able to define a Grafana dashboard based on their workload metrics, and have that provisioned upon relation with COS. The current PR exposes the metrics on a different port, which is the first step. The next step is adding logic for the charm to pick up the user-defined Grafana dashboard from a specific place in the workload image, and inject it to where it needs to be such that it is picked up by the COS relation. The expected result is that upon relating to COS, we will see two Grafana dashboards: one for the charm metrics, and one for the workload metrics. This needs to also be shown via some sort of screenshot attached to the PR for validation.

yakupt-canonical:
> I believe this requires more thought on the requirements part. What is the end goal here? An operator should be able to define a Grafana dashboard based on their workload metrics, and have that provisioned upon relation with COS. The current PR exposes the metrics on a different port, which is the first step. The next step is adding logic for the charm to pick up the user-defined Grafana dashboard from a specific place in the workload image, and inject it to where it needs to be such that it is picked up by the COS relation. The expected result is that upon relating to COS, we will see two Grafana dashboards: one for the charm metrics, and one for the workload metrics. This needs to also be shown via some sort of screenshot attached to the PR for validation.

@kelkawi-a Great suggestion, let me check the grafana dash provider docs for how to handle that.

config.yaml
yakupt-canonical: I tried unsetting (`juju config --reset workload-prometheus-port`) it and it worked but let me verify again.

gaborbk: Thanks, great that it works like that - let's also make it clear in the charm spec though

gaborbk:
> > I believe this requires more thought on the requirements part. What is the end goal here? An operator should be able to define a Grafana dashboard based on their workload metrics, and have that provisioned upon relation with COS. The current PR exposes the metrics on a different port, which is the first step. The next step is adding logic for the charm to pick up the user-defined Grafana dashboard from a specific place in the workload image, and inject it to where it needs to be such that it is picked up by the COS relation. The expected result is that upon relating to COS, we will see two Grafana dashboards: one for the charm metrics, and one for the workload metrics. This needs to also be shown via some sort of screenshot attached to the PR for validation.
> 
> @kelkawi-a Great suggestion, let me check the grafana dash provider docs for how to handle that.

Thanks both, I agree we should have the Grafana dashboards propagate as well. And would be nice to check if all logs are collected by the existing setup or we need to add something for that too.

gaborbk:
> @kelkawi-a @gaborbk It seems that `GrafanaDashboardProvider` expects user-defined dashboards to be stored under `charm` directory:
> 
> https://github.com/canonical/grafana-k8s-operator/blob/0a91c1528ec98cbcefe74bbc84e8b74aff0e410b/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L1154-L1157
> 
> > **dashboards_path**: a filesystem path **relative to the charm root**
> > where dashboard templates can be located. By default, the library
> > expects dashboard files to be in the `<charm-py-directory>/grafana_dashboards`
> > directory.
> 
> https://github.com/canonical/grafana-k8s-operator/blob/0a91c1528ec98cbcefe74bbc84e8b74aff0e410b/lib/charms/grafana_k8s/v0/grafana_dashboard.py#L1164
> 
> ```python
> dashboards_path = _resolve_dir_against_charm_path(charm, dashboards_path)
> ```
> 
> This means that dashboards must live in `charm` pod. Since my knowledge on charms and rocks is quite limited, could you please point me a direction if it is possible to patch / add contents to charm pod when using rocks.
> 
> If it is not possible or easy, I would request you to review the current state of the PR, because it would be still useful to have metrics endpoint since people can define dashboards on Grafana themselves.

You're right @yakupt-canonical , the Grafana integration automatically picks up dashboard definitions from the `grafana_dashboards` directory under the charm - see https://github.com/canonical/contracts/tree/main/charm/src/grafana_dashboards . The running OCI image can and should not modify the charm that runs it, although some deault workload-agnostic dashboards can live there - just like they do now. For workload-specific dashboards, I think we need to have them defined outside this charm, for example using `cos-configuration`. Specifically for Contracts thisis perfectly fine, we want to move all our dashboard and alert rule definitions there soon anyway. For other users of this charm I think we just need to document how this works, WDYT @kelkawi-a ?

yakupt-canonical:
@kelkawi-a @gaborbk I updated the docs to reflect new workload metrics config and relation, please review.

README.md
```diff
@@ -286,6 +286,23 @@ juju run grafana/0 -m cos get-admin-password --wait 1m
...
+If your workload exports metrics, then it can also be related to
+[Canonical Observability Stack](https://charmhub.io/topics/canonical-observability-stack).
+
+To enable workload metrics, run the following command:
+
+```bash
+juju config temporal-worker-k8s workload-prometheus-port <workload-metrics-port>
+```
+
+Then run the following command to relate it to cos-lite:
```
gaborbk, line 300:
```suggestion
Then run the following command to relate it to cos:

```

README.md
```diff
@@ -286,6 +286,23 @@ juju run grafana/0 -m cos get-admin-password --wait 1m
...
+```bash
+# Relate workload metrics cos-lite:
```
gaborbk, line 302:
```suggestion
```
Doesn't seem to add much value, we already say what this line is doing above

gaborbk (approved):
LGTM with nits

src/charm.py
```diff
@@ -81,6 +81,15 @@ def __init__(self, *args):
             refresh_event=self.on.config_changed,
         )
 
+        if self.config.get("workload-prometheus-port"):
```
kelkawi-a, line 84: The original integration test suite does not cover the COS relation. However, I believe we would benefit from an integration test here to validate this feature. This would entail modifying the `resource_sample_py` to expose some simple metric, that can be queried via an integration test.

kelkawi-a:
> > The idea is for each OCI image to have a set of defined Grafana dashboard(s) and alert rules that can be injected into the charm at runtime and be created upon relation to COS. This PR does the first part of exposing workload metrics.
> 
> @kelkawi-a My experience / knowledge on charms and OCI images are quite limited, if you can point me some documentation or tutorial, I can try adding that or we can do it as a follow up PR if it would take some time.

[This PR](https://github.com/canonical/temporal-worker-k8s-operator/pull/32/files#diff-7dbebe947504f6e44c312cf093f44032d41ecb48d37b289d1da7334069bbee67) can be used as a reference on how to modify the sample resource and add a corresponding integration test. I'd be happy to jump on a call and discuss this further. Merging the PR without a test is not recommended as there is no validation on this feature being added to the charm.

src/charm.py
yakupt-canonical: I added the integration test, thank you

yakupt-canonical:
@kelkawi-a @gaborbk I added the integration test, please review again, thank you.

tests/integration/test_workload_metrics.py
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env python3
+# Copyright 2023 Canonical Ltd.
```
gaborbk, line 2:
```suggestion
# Copyright 2025 Canonical Ltd.
```

gaborbk (approved):
Awesome, thanks! LGTM with a tiny nit

resource_sample_py/resource_sample/activities/workload_metrics.py
```diff
@@ -0,0 +1,16 @@
+# Copyright 2023 Canonical Ltd.
```
yakupt-canonical, line 1:
```suggestion
# Copyright 2025 Canonical Ltd.
```

tests/integration/test_workload_metrics.py
yakupt-canonical: I had to revert this because license check lint failed with this change.

gaborbk: Hmm, strange, I see that the copyright year is set to 2023 in https://github.com/canonical/temporal-worker-k8s-operator/blob/main/.licenserc.yaml . If we update that file, we need to also update every place in this repo that says 2023. @kelkawi-a what's you preference here, should we just leave the year as 2023?

yakupt-canonical: @gabor I tried setting the year to 2023-2025 hoping it would treat as a list of years but no, the tool is not that smart, it only does a string replace with what is written. So either we will update all of them 2025 or leave it as 2023.

gaborbk: Yep, that's really unfortunate :( I wonder if we could simply disable the year validation?

kelkawi-a: Let's leave it at 2023 for now as that's the year the repo was created. This repo (along with other Temporal-related ones) will live with the Data Workflows team, so they can make the call when needed.

tests/integration/test_workload_metrics.py
```diff
@@ -0,0 +1,56 @@
...
+    time.sleep(70)
+    prometheus_url = await get_unit_url(ops_test, "prometheus-k8s", 0, 9090, "http")
+    with requests.Session() as http:
+        query_url = f"{prometheus_url}/api/v1/query?query=activity_execution_count"
+        resp = http.get(query_url)
+
+    assert resp.status_code == 200
+
+    # if configured correctly there should be at least one metric present.
+    assert resp.text.count("activity_execution_count") > 0
```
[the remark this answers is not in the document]
yakupt-canonical: @kelkawi-a Please re-review, I clarified the test by changing the metric name.

kelkawi-a (approved):
LGTM, thank you for working on this @yakupt-canonical ! I will invite @shayancanonical to take a look at this as well, since his team now owns the development of the Temporal charms.

Shayan, this PR does one part of what we discussed in Frankfurt, which is introduce the ability of the workload container to export metrics that can be consumed by COS. The next part would be to allow workloads to also define custom Grafana dashboards and alerts which would be picked up by the charm and also propagated to COS. Please take a look.

kelkawi-a:
cc @DnPlas

src/charm.py
```diff
@@ -81,6 +81,15 @@ def __init__(self, *args):
...
+            workload_prometheus_port = self.config.get("workload-prometheus-port")
+            self._app_prometheus_scraping = MetricsEndpointProvider(
+                self,
+                relation_name="workload-metrics-endpoint",
+                jobs=[{"static_configs": [{"targets": [f"*:{workload_prometheus_port}"]}]}],
+                refresh_event=self.on.config_changed,
+            )
+
```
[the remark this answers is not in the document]
DnPlas: Thanks for the reply @sed-i ! Yeh, at first I thought of putting both targets in the same job, but I wanted to have a clear separation between the charm and the workflow...I guess to cluster the metrics for a better UX.

config.yaml
```diff
@@ -158,30 +158,34 @@ options:
...
+    description: Port of the workload defined metrics endpoint.
```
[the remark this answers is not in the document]
shayancanonical: I agree with @sed-i: there does not seem to be a compelling reason for us to allow the operator to configure the port of the metrics endpoint - it seems like a technical implementation detail (since I think that generally, the config options should allow the operator to configure behavior of the charm - the ability to change the port does not really feel a natural config option)

config.yaml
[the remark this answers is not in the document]
shayancanonical: Not entirely sure that the SDK is able to specify the URL path for the metrics endpoint. https://python.temporal.io/temporalio.runtime.PrometheusConfig.html seems to indicate that the address is configurable, but unsure if the path is considered part of the address

config.yaml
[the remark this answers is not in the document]
DnPlas: Not providing the config option means we'll need to tell users which port to use in their worker configurations, right? Whenever they create a worker (e.g., worker.py or the Go worker binary), the metrics port will always default to whatever is defined in the charm.

Does this align to what you were thinking @shayancanonical ?

---
#### Context

According to Temporal docs, we need to "set a Prometheus endpoint before any other Temporal code.". A Python example would be:

```python
from temporalio.runtime import Runtime, TelemetryConfig, PrometheusConfig

# Create a new runtime that has telemetry enabled. Create this first to avoid
# the default Runtime from being lazily created.
new_runtime = Runtime(telemetry=TelemetryConfig(metrics=PrometheusConfig(bind_address="0.0.0.0:9000")))
my_client = await Client.connect("my.temporal.host:7233", runtime=new_runtime)
```

In this case, the `9000` would be the port that this charm config option is referring to. This is port is configured in the worker - which is a process running in the sidecar container of the `temporal-worker-k8s` charm. This will require us to make sure the metrics port from the worker is actually exposed (e.g. with `self.unit.set_ports` or the `KubernetesServicePatch`). This is where the charm config option would matter since it will take any port and make sure it is setup.

FTR, I'm not against removing the config option, just stating how it would be used.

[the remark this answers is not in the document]
shayancanonical: I had not considered the workflow code changing the port in the worker, but now that you mention it, I am wondering if there's a clean way for the worker charm code to pick up which port the workflow is exposing metrics on automatically (and not have the necessity for the operator to update the config option). Wdyt?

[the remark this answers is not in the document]
yakupt-canonical: One of the reason for having the config was to be optionally enable the metrics-endpoint relation, if user provides a config for the port, the `workflow-metrics-endpoint` relation is exposed, otherwise there would be no relation.

I don't have enough experience on charms, but in the current state, any existing workflow can be deployed without any modification with the worker operator, but in order charm code to pick up the port, I think we need to introduce some coupling between workflow code and temporal worker charm, though I may be wrong.

If there is a clean way of doing it, I am happy to implement it.

src/charm.py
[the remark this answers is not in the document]
yakupt-canonical: I updated the code with your suggestion, thank you, PTAL.

tests/integration/test_workload_metrics.py
```diff
@@ -0,0 +1,56 @@
...
+        apps=[APP_NAME, "prometheus-k8s"],
+        status="active",
+        raise_on_blocked=False,
+        timeout=600,
+    )
+
+    await run_sample_workflow(ops_test, workflow_type="workload_metrics")
+    # The scrape interval is 1m, so wait some time to make sure that
+    # prometheus scrapes the metrics.
+    time.sleep(70)
```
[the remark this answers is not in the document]
yakupt-canonical: I updated the test with tenacity retries, thank you for the suggestion, PTAL.

yakupt-canonical:
@DnPlas @shayancanonical @kelkawi-a I updated the PR with the suggested changes, please review.

PR #42, opened 2025-06-05
[WF-24] Convert charm related unit tests from harness to use scenario

shayancanonical:
The project is using the deprecated harness framework to implement unit tests. This PR implements existing charm related unit tests with scenario instead and deletes the harness tests. The unit tests for the `State` abstraction are not ported over since we are planning to refactor out the usage of this abstraction in the near future.

In order to be able to use the latest scenario/ops.testing extra, we had to update the `vault_kv` charmlib hosted by `vault_k8s` from v0.12 to v0.16 and update the `ops` package used by the project.


lib/charms/vault_k8s/v0/vault_kv.py
DnPlas, line 1: What was the need for bumping this lib?

DnPlas (commented):
Thanks @shayancanonical ! LGTM, just a little question.

lib/charms/vault_k8s/v0/vault_kv.py
shayancanonical: This lib was bumped for us to be able to upgrade `ops` to a version that has the latest `ops[testing]` python package extra

It's actually a convoluted rabbit hole: the `vault_kv` charm lib imports the `interface_tester` python package (which confirms that a charm lib is adhering to a repository where charm lib interfaces are defined). The `interface_tester` package imported in v12 of the `vault_kv` lib depends on an old version of `ops-scenario` -> to be able to have a newer `ops[testing]` version and avoid version conflicts, we needed to bump the `vault_kv` lib to v16

DnPlas: Got it, thanks for the explanation. Ideally we'll have separate PRs for these sort of "unrealted" changes, but for this one I agree. No action required.

PR #43, opened 2025-06-11
Run integration tests against 3.6/stable juju

shayancanonical:
We would like to start testing all temporal charms against the LTS juju version (3.6/stable)


DnPlas (approved):
Thanks @shayancanonical !

PR #70, opened 2025-09-09
[WF-113] Add terraform module for temporal-worker

shayancanonical:
## Description

Adds terraform module for temporal-worker-k8s
Adds CI to format, lint and apply terraform module as a sanity check

---
## Testing instructions

terraform fmt terraform/
terraform init
juju add-model terraform
terraform apply  -var 'image={"image": "ghcr.io/canonical/temporal-worker:e2524cf20e20b00369a15052e849f595ac2331cd-_1.0_amd64"}' -var 'model=terraform'

---
## Checklist (all that apply)
- [x] Integration tests added or updated
- [x] Documentation updated

### DIFF
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -21,25 +21,17 @@ jobs:
         run: |
           sudo snap install terraform --classic
 
-      - name: Initialize Terraform module
-        run: terraform init
-
-      - name: Validate Terraform Module
-        run: terraform validate -no-color
-
-      - name: Validate Terraform fmt
+      - name: Run terraform format
         run: |
-          set +e
-          terraform fmt -recursive -check -diff
-          FMT_STATUS="$?"
-          if [[ "$FMT_STATUS" -ne 0 ]]; then
-            echo "❌ terraform fmt failed" >> "$GITHUB_STEP_SUMMARY"
-          fi
-          exit "$FMT_STATUS"
+          just validate
+          just fmt
 
   lint:
     name: Terraform lint
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./terraform
 
     steps:
       - name: Checkout repository
@@ -52,11 +44,14 @@ jobs:
           sudo snap install tflint
 
       - name: Run tflint
-        run: tflint --chdir ./terraform/
+        run: just lint
 
   apply:
     name: Terraform apply
     runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./terraform
 
     steps:
       - name: Checkout repository
@@ -73,29 +68,9 @@ jobs:
           sudo snap install concierge --classic
           sudo concierge prepare -p k8s
 
-      - name: Add juju model
-        run: juju add-model terraform
-
       - name: Install dependencies
         run: |
           sudo snap install terraform --classic
 
-      - name: Apply terraform
-        working-directory: terraform
-        run: |
-          set +e
-
-          terraform init
-
-          terraform apply -auto-approve -var 'image={"image": "ghcr.io/canonical/temporal-worker:e2524cf20e20b00369a15052e849f595ac2331cd-_1.0_amd64"}' -var 'model=terraform' -var 'channel=latest/edge'
-
-      - name: Wait for application
-        run: |
-          juju wait-for model terraform --query="forEach(units, unit => (unit.workload-status == 'blocked'))"
-      
-      - name: Juju status
-        run: juju status
-
-      - name: Dump logs
-        if: failure()
-        uses: canonical/charming-actions/dump-logs@2.6.3
+      - name: Run terraform apply
+        run: just test
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -1,0 +1,87 @@
+name: Terraform checks
+
+on:
+  pull_request:
+
+jobs:
+  format:
+    name: Terraform format
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: ./terraform
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.ref }}
+
+      - name: Setup Terraform
+        uses: hashicorp/setup-terraform@v3
+
+      - name: Initialize Terraform module
+        run: terraform init
+
+      - name: Validate Terraform Module
+        run: terraform validate -no-color
+
+      - name: Validate Terraform fmt
+        run: |
+          set +e
+          terraform fmt -recursive -check -diff
+          FMT_STATUS="$?"
+          if [[ "$FMT_STATUS" -ne 0 ]]; then
+            echo "❌ terraform fmt failed" >> "$GITHUB_STEP_SUMMARY"
+          fi
+          exit "$FMT_STATUS"
+
+  lint:
+    name: Terraform lint
+    uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-lint.yaml
+    with:
+      module-path: .
+
+  apply:
+    name: Terraform apply
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up operator environment
+        run: |
+          set +e
+          sudo snap install concierge --classic
+          sudo concierge prepare -p k8s
+
+      - name: Add juju model
+        run: |
+          set +e
+          juju bootstrap k8s k8s
+          juju add-model terraform
+
+      - name: Install dependencies
+        run: |
+          sudo snap install terraform --classic
+
+      - name: Apply terraform
+        working-directory: terraform
+        run: |
+          set +e
+
+          terraform init
+
+          terraform apply -var 'image={"image": "ghcr.io/canonical/temporal-worker:e2524cf20e20b00369a15052e849f595ac2331cd-_1.0_amd64"}' -var 'model=terraform'
+
+      - name: Wait for application
+        run: |
+          juju wait-for model terraform --query="forEach(units, unit => (unit.workload-status == 'blocked'))"
+      
+      - name: Juju status
+        run: juju status
+
+      - name: Dump logs
+        if: failure()
+        uses: canonical/charming-actions/dump-logs@2.6.3
--- a/.github/workflows/terraform.yaml
+++ b/.github/workflows/terraform.yaml
@@ -18,7 +18,8 @@ jobs:
           ref: ${{ github.ref }}
 
       - name: Setup Terraform
-        uses: hashicorp/setup-terraform@v3
+        run: |
+          sudo snap install terraform --classic
 
       - name: Initialize Terraform module
         run: terraform init
@@ -38,9 +39,20 @@ jobs:
 
   lint:
     name: Terraform lint
-    uses: canonical/charmed-kubeflow-workflows/.github/workflows/terraform-lint.yaml@main
-    with:
-      module-path: .
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Install dependencies
+        run: |
+          set +e
+          sudo snap install terraform --classic
+          sudo snap install tflint
+
+      - name: Run tflint
+        run: tflint --chdir ./terraform/
 
   apply:
     name: Terraform apply
@@ -55,6 +67,7 @@ jobs:
           set +e
           sudo snap install concierge --classic
           sudo concierge prepare -p k8s
+          sudo k8s bootstrap
 
       - name: Add juju model
         run: |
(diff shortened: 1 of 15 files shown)


.github/workflows/terraform.yaml
```diff
@@ -0,0 +1,101 @@
...
+          sudo snap install terraform --classic
+
+      - name: Apply terraform
+        working-directory: terraform
+        run: |
+          set +e
+
+          terraform init
+
+          terraform apply -auto-approve -var 'image={"image": "ghcr.io/canonical/temporal-worker:e2524cf20e20b00369a15052e849f595ac2331cd-_1.0_amd64"}' -var 'model=terraform'
```
DnPlas: I'm not entirely in favour of:

1. Using this image for testing as it is not being published any more, meaning it will get outdated; on the other hand, I understand we just need an image that has the `./app/scripts/start-worker.sh` entrypoint.
2. Leaving this image inside this workflow - I'd like this workflow to be reusable rather than a copy paste, meaning this "hardcoded" values will have to go away.

.github/workflows/terraform.yaml
DnPlas, line 1: Did you consider just creating a reusable workflow? It can be hosted at canonical/workflows-team or canonical/workflows-team-workflows.

I strongly suggest we do it since we are going to be using this for at least four repos and potentially on the solution repo.

terraform/outputs.tf
```diff
@@ -0,0 +1,18 @@
+output "name" {
```
DnPlas: According to [this](https://docs.google.com/document/d/1EG71A2pJ244PQRaGVzGj7Mx2B_bzE4U_OSqx4eeVI1E/edit?tab=t.0) spec, it should be `app_name`.

terraform/variables.tf
```diff
@@ -0,0 +1,124 @@
+variable "name" {
```
DnPlas:
```suggestion
variable "app_name" {
```

terraform/variables.tf
```diff
@@ -0,0 +1,124 @@
...
+variable "revision" {
+  type        = number
+  description = "Revision of the charm to deploy"
+  default     = 23
+}
+
+variable "channel" {
+  type        = string
+  description = "Charmhub channel to deploy the charm from"
+  default     = "latest/edge" # TODO: change to 1.0/edge
```
DnPlas: Is this because we haven't been able to release the worker charm? If yes, please link here the GH issue so we know when this should be changed, otherwise, the terraform module is only supported for 1.0/edge, I think we can just make the change in this PR.

terraform/variables.tf
DnPlas, line 1: I see a number of configurations defined here, but according to the spec, they should rather be something like:

```
variable "config" {
  description = "Map of charm configuration options"
  type        = map(string)
  default     = {}
}
```

Is there a reason to deviate from the spec?

terraform/versions.tf
```diff
@@ -0,0 +1,14 @@
...
+      source  = "juju/juju"
+      version = ">= 0.21.1"
+    }
+
+    null = {
+      source  = "null"
+      version = ">=3.2.4"
+    }
+  }
+}
```
DnPlas, line 14: I'm unsure what this is used for.

tox.ini
```diff
@@ -42,13 +42,13 @@ deps =
...
-    flake8==5.0.4
+    flake8==7.0.0
     flake8-builtins==1.5.3
-    flake8-copyright==0.2.3
+    flake8-copyright==0.2.4
     flake8-docstrings==1.6.0
     isort==5.10.1
     pep8-naming==0.13.2
-    pyproject-flake8==5.0.4.post1
+    pyproject-flake8==7.0.0
```
DnPlas: This change was introduced by https://github.com/canonical/temporal-worker-k8s-operator/pull/72, no need to have it here.

terraform/main.tf
```diff
@@ -0,0 +1,61 @@
...
+    sentry-environment   = var.sentry_config["environment"]
+    sentry-redact-params = var.sentry_config["redact_params"]
+    sentry-sample-rate   = var.sentry_config["sample_rate"]
+
+    auth-secret-id = var.auth_secret_id
+
+    tls-root-cas = var.tls.root_cas
+
+    db-name = var.db_name
+  }
```
DnPlas: In various instances of charm modules I've seen the config just defined as:

```
resource "juju_application" "alertmanager" {
  name               = var.app_name
  config             = var.config
  ```
  
  but we need variables.tf to have a config as a map (see my other comment). I believe the alertmanager approach aligns more to what the spec says. I also see it [here](https://github.com/canonical/kfp-operators/tree/main/charms/kfp-api/terraform).

File: terraform/variables.tf [hunk 8]
```diff
@@ -0,0 +1,124 @@
...
+}
+
+variable "image" {
+  type = object({
+    image    = string
+    username = optional(string)
+    password = optional(string)
+  })
+  description = "Details of the worker image"
+}
```
DnPlas: I imagine username and password are the registry's and not the image's(?). Should we change the name to make that more clear or include that information in the description?

terraform/main.tf
```diff
@@ -0,0 +1,61 @@
...
+    command = <<EOT
+      echo "${yamlencode({
+    "registrypath" = var.image.image,
+    "username"     = var.image.username,
+    "password"     = var.image.password,
+})}" > temporal_worker_image.yaml
+
+      juju attach-resource ${var.name} temporal-worker-image=temporal_worker_image.yaml
+
+      rm temporal_worker_image.yaml
```
DnPlas: I may be missing some information here, but how will our charm read the contents of the `temporal_worker_image.yaml` file and when will it refresh the image?

DnPlas (commented):
Thanks @shayancanonical ! Left a couple comments, please also add a README.md (similar to [this](https://github.com/canonical/kfp-operators/blob/main/charms/kfp-api/terraform/README.md))

terraform/outputs.tf
```diff
@@ -0,0 +1,18 @@
...
+  }
+}
+
+output "requires" {
+  value = {
+    logging  = "logging"
+    vault    = "vault"
+    database = "postgresql_client"
+  }
+}
```
DnPlas, line 18: You could also leave them as `endpoints`, as suggested in the spec, but I'm okay with this as well. Not a blocker.

.github/workflows/terraform.yaml
shayancanonical: I agree. I'm currently working through a flexible workflow for testing terraform modules for the workflows team (to be used in the individual charm modules + in the solution repo). I will create a ticket as a follow up as I'd rather iterate on this item instead of blocking this PR until the workflow is established. This will also unblock the implementation of the solution module

Furthermore, I am also considering of adding a simple Dockerfile that will be used to generate a simple image to be used for CI testing. I am also very inclined towards using `just` to encapsulate these builds so the CI is simple and leaves us open to configuring builds in individual repos

.github/workflows/terraform.yaml
shayancanonical: I agree. https://github.com/canonical/temporal-worker-k8s-operator/pull/70#discussion_r2348823213

Please review this PR just on the basis of functionality. I will iterate to abstract and simplify the testing components as well as image generation

terraform/outputs.tf
shayancanonical: addressed in c187aed65a44f17337a086e03e962bef67774bd3

terraform/variables.tf
shayancanonical: addressed in 899c0d205a6691845d18641e90b37e990d90e8ab

terraform/variables.tf
shayancanonical: Set default channel to `1.0/stable` in cfe2a8cd414b5ebb2e58d04f76ac7b7ef0ad14e0 (even though the channel is not yet published

Will use the tf var to overwrite this to `latest/edge` in CI temporarily

terraform/variables.tf
shayancanonical: Addressed in 899c0d205a6691845d18641e90b37e990d90e8ab

terraform/versions.tf
shayancanonical: This is the provider for the `null_resource` used in main.tf (the null resource is used to ensure that we are able to run `juju attach-resource` with images from a private image repository)

terraform/main.tf
shayancanonical: addressed 899c0d205a6691845d18641e90b37e990d90e8ab

terraform/variables.tf
shayancanonical: Renamed to `registry_username` and `registry_password` to eliminate ambiguity

terraform/main.tf
shayancanonical: This is actually the biggest pain of this module. The juju provider [does not](https://registry.terraform.io/providers/juju/juju/latest/docs/resources/application#resources-1) only allows URLs or revision numbers to be used as OCI image resources. There's no way to provide additional details like `username` or `password` to the provider

The way it works now is that the app will be deployed with the image attached to the charm revision that is deployed. When we run `juju attach-resource` the app gets refreshed with the specified image in vars

What will be our strategy for deploying temporal-worker-k8s without an image in the near future? This may break if we do not upload an OCI image resource with newly released revisions in the future

terraform/outputs.tf
shayancanonical: Please advise - I'm ok with both approaches

The spec defines `endpoints` as an output, but I think it's rather nice to be able to distinguish between the requires and provides endpoints

.github/workflows/terraform.yaml
shayancanonical: Added simple dockerfile that builds a compatible image with a script that busy waits. This can be used to validate that the module deploys the application into blockedstatus successfully

.github/workflows/terraform.yaml
shayancanonical: added a reusable workflow, starting https://github.com/canonical/temporal-worker-k8s-operator/pull/70/commits/1a17934310fead84732f5d27572395e9086efa41

shayancanonical:
Added a README for terraform module in https://github.com/canonical/temporal-worker-k8s-operator/pull/70/commits/8295977750670c9f8c23bf8137bf57c530487036

terraform/main.tf
shayancanonical: https://github.com/juju/terraform-provider-juju/issues/620 tracks this issue

The result of this issue being addressed should be the support for passing registry credentials directly to terraform provider. Sadly, I can't think of a cleaner way to accomplish the intended functionality without the above work around

terraform/outputs.tf
DnPlas: I also agree that having a distinction between requires/provides can help with usability. Since the spec just suggests endpoints, we can leave this as is.

terraform/main.tf
DnPlas: > What will be our strategy for deploying temporal-worker-k8s without an image in the near future? This may break if we do not upload an OCI image resource with newly released revisions in the future

I'm not sure I fully understand this question. May I ask for clarification?

> The result of this issue being addressed should be the support for passing registry credentials directly to terraform provider. Sadly, I can't think of a cleaner way to accomplish the intended functionality without the above work around

I'm okay with the workaround, I was trying to understand it better rather than question it.

terraform/test/test.tfvars
```diff
@@ -0,0 +1,3 @@
+model   = "terraform"
+image   = { "image" : "docker://localhost:5000/temporal-worker:test-terraform" }
+channel = "latest/edge" # TODO: update to 1.0/edge once channel published
```
DnPlas: Let's create an issue to track this, you can add it to the same epic. Please link it here as a comment.

terraform/main.tf
DnPlas: Let's add a comment in this and in the var file to make sure we remind ourselves that this may be changed in the near future (when https://github.com/juju/terraform-provider-juju/issues/620 gets resolved).

terraform/justfile
```diff
@@ -0,0 +1,68 @@
...
+    #!/usr/bin/bash
+    set -euxo pipefail
+
+    trap 'just destroy "./test/test.tfvars"' EXIT
+
+    just add-model
+
+    just apply "./test/test.tfvars"
+
+    juju wait-for model terraform --query="forEach(units, unit => (unit.workload-status == 'blocked'))"
```
DnPlas: Is it possible to check that it is blocked because of an expected reason?

terraform/main.tf
shayancanonical: added comment in 28121b232bdb1ac9731f7be3c2fb9368602b10df

terraform/test/test.tfvars
shayancanonical: added issue and added link in 28121b232bdb1ac9731f7be3c2fb9368602b10df

terraform/justfile
shayancanonical: checking workload-message as well in 28121b232bdb1ac9731f7be3c2fb9368602b10df

shayancanonical:
Changed Dockerfile to rockcraft for the test image to be used in terraform tests (28121b232bdb1ac9731f7be3c2fb9368602b10df)

DnPlas (dismissed):
Thanks @shayancanonical ! Looks really nice.

DnPlas (approved):
Re-approving after the changes required for using re-usable workflow. Thanks @shayancanonical !

PR #86, opened 2025-12-04
[WF-334] Add temporal-worker-consumer and temporal-host-info relations

wyattrees:
## Description
Adds handling for relations `temporal-host-info` and `temporal-worker-consumer` 

---
## Testing instructions
For `temporal-host-info`
1. Build charm
2. Deploy temporal-k8s
3. Ensure `host` config option is empty
4. integrate via `temporal-host-info` interface
5. Validate that worker can connect to the server

For `temporal-worker-consumer`
1. Build charm
6. Build test charm (tests/integration/worker_consumer_requirer)
7. Deploy charms
8. Integrate with test charm
9. Check `juju status` workload message for test charm shows the namespace and queue configured in `temporal-worker-k8s`

This keeps the `host` config option and allows it to take precedence over the `temporal-host-info` relation data as to not break existing deployments that upgrade to this version.

---
## Checklist (all that apply)
- [X] Unit tests added or updated
- [ ] Documentation updated


lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
+# Copyright 2025 Canonical Ltd.
```
shayancanonical: nit: update copyright years to 2026 - to reflect the year this file will be introduced in the default branch

lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
...
+
+from ops import (
+    ConfigChangedEvent,
+    Handle,
+    RelationChangedEvent,
+    RelationJoinedEvent,
+    framework,
+)
+from ops.charm import CharmBase
+from ops.framework import EventBase, EventSource, ObjectEvents
```
shayancanonical: we can eliminate `ops.framework` since all the classes in framework can be directly accessed from the parent `ops` module

lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
+# Copyright 2025 Canonical Ltd.
+# See LICENSE file for licensing details.
+
```
shayancanonical: we need to add a docstring for this charm lib to explain how the classes in this lib will be used by charms that import it

lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
...
+    @property
+    def relations(self) -> list[Relation]:
+        """Return the relations for this interface."""
+        return self.charm.model.relations.get(RELATION_NAME, [])
+
+    @property
+    def namespace(self) -> str | None:
+        """Return the namespace from the relation data."""
+        for relation in self.relations:
```
shayancanonical: this implies that there may be more than one upstream relation (since `charm.model.relations.get(RELATION_NAME)` may include more than 1 relation

we should:
1. either question if a downstream app (that triggers workflows) can be related to multiple temporal-worker charms (i dont see why not)
2. if so, then we need to ensure that we are providing the namespace for the right relation (so take the relation_id as a kwarg and return the namespace/queue for that relation)
3. we can also add a property that returns a mapping of relation_id to namespace and queue for a full traversal of all upstream workers

lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
...
+
+        :param: event: The relation event that triggered this handler.
+        :type event: RelationChangedEvent | RelationJoinedEvent
+        """
+        try:
+            namespace = event.relation.data[event.relation.app]["namespace"]
+            queue = event.relation.data[event.relation.app]["queue"]
+        except KeyError:
+            self.charm.unit.status = WaitingStatus("Waiting for temporal-worker-consumer provider")
+            event.defer()
```
shayancanonical: do not think we need to defer here, as another relation change is required for upstream to update relation databags which will fire another relation-changed event

lib/charms/temporal_worker_k8s/v0/temporal_worker_consumer.py
```diff
@@ -0,0 +1,191 @@
...
+            return
+        self.charm.unit.status = ActiveStatus()
```
shayancanonical: i am strictly against the charm lib setting any statuses on behalf of the charm - because it makes tracing of charm code much more complex 

would suggest to instead expose a `is_ready()` method that can take an optional kwarg for relation_id. this method returns if all expected data for the relation_id exists, or if relation_id=None, then if all upstream relations have the expected data

src/charm.py
```diff
@@ -353,19 +362,38 @@ def _update(self, event):  # noqa: C901
             if value:
                 context.update({key: value})
 
+        host = None
+        if self.config.get("host"):
```
shayancanonical: same as adjacent PRs, would recommend truly deprecating the `host` config and moving to `BlockedStatus` if set

shayancanonical (commented):
leaving partial review since raising concerns about the `temporal_worker_consumer` charm lib

additionally, we should question if there's a better name for the charm lib (e.g. `temporal_worker.py`) and the relation interface (e.g. `temporal-worker`)

lib/charms/temporal_worker_k8s/v0/temporal_worker_info.py
```diff
@@ -0,0 +1,280 @@
...
+
+    def _on_config_changed(self, event: ConfigChangedEvent) -> None:
+        """Handle config changes by updating relation data.
+
+        Args:
+            event: The config changed event that triggered this handler.
+        """
+        logger.info("Config changed, updating temporal-worker-info relation data")
+        if self.charm.unit.is_leader():
+            for relation in self.charm.model.relations.get(RELATION_NAME, ()):
```
shayancanonical: (non-blocking, nit):  while it works, the return type of `self.charm.model.relations.get()` would be a list, so the default value should be `[]` instead of `()`

lib/charms/temporal_worker_k8s/v0/temporal_worker_info.py
```diff
@@ -0,0 +1,280 @@
...
+            ``(None, None)`` if the relation is missing or incomplete.
+        """
+        rel = self.charm.model.get_relation(RELATION_NAME, relation_id)
+        if not rel or not rel.app:
+            return None, None
+        data = rel.data[rel.app]
+        return data.get("namespace"), data.get("queue")
+
+    @property
+    def namespace(self) -> str | None:
```
shayancanonical, line 239: nit: can we leverage `relation_payloads` or `get_namespace_queue` to implement these properties? this way, the logic that encapsulates the check for a valid `namespace` and valid `queue` is implemented in one place

or better yet, we can implement a private method `_get_valid_relation_data(relation_id)` that extracts the namespace and queue of a relation, and can be used by almost all the getters in this class (even `is_ready()`)

if we add another required field in the relation in the future, it'll be easier to maintain and test

lib/charms/temporal_worker_k8s/v0/temporal_worker_info.py
```diff
@@ -0,0 +1,280 @@
...
+        """Emit :attr:`on.temporal_worker_info_available` when this relation is complete.
+
+        Args:
+            event: Relation joined or changed event for ``temporal-worker-info``.
+        """
+        if not self.is_ready(event.relation.id):
+            return
+        app = event.relation.app
+        namespace = event.relation.data[app]["namespace"]
+        queue = event.relation.data[app]["queue"]
```
shayancanonical: (non-blocking, nit): these variables are not multi-use, they're only accessed once. i'd prefer to just inline them if so (imo, this results in more readable code)

tests/integration/worker_info_requirer/requirements.txt
```diff
@@ -0,0 +1 @@
+ops==2.21.1
```
shayancanonical: can we avoid creating a `requirements.txt` file if the directory already has a `pyproject.toml`?

tests/integration/helpers.py
```diff
@@ -173,23 +173,32 @@ async def run_sample_workflow(ops_test: OpsTest, workflow_type=None, use_env_var
...
-        .run_action("tctl", args="--ns default namespace register -rd 3")
+        .run_action("tctl", args=f"--ns {namespace} namespace register -rd 3")
     )
     result = (await action.wait()).results
-    logger.info(f"tctl result: {result}")
+    logger.info("tctl namespace %s: %s", namespace, result)
     assert "result" in result and result["result"] == "command succeeded"
 
 
+async def create_default_namespace(ops_test: OpsTest):
```
shayancanonical: suggest we eliminate this helper function as it is solely a wrapper around the `register_temporal_namespace` and just call `register_temporal_namespace(ops_test, "default")` instead?

shayancanonical (commented):
looks great! left some nits + opportunities to make the code more concise. ready to approve once you've had a chance to respond

shayancanonical (commented):
changes look great! ready to approve PR once integration tests pass

terraform/justfile
```diff
@@ -16,10 +16,27 @@ start-local-registry: (stop-local-registry)
 [private]
 build-and-push-image: (start-local-registry)
     #!/usr/bin/bash
+    set -euxo pipefail
     cd test
-    rockcraft pack
-
-    rock_file=$(ls | grep "\.rock")
+    for attempt in {1..3}; do
```
shayancanonical, line 21: why are we retrying pack of the rock?

shayancanonical (approved):
changes look good

left a question about a change in the terraform justfile. please triage and resolve open comments from previous rounds of feedback. happy to help close the comments out once you've had a chance to respond whether the comments have been addressed in subsequent commits (comments must be resolved before PR merge is enabled)

lib/charms/temporal_worker_k8s/v0/temporal_worker_info.py
nikita-jare: Thanks for flagging this. I started by reusing `get_namespace_queue` in the `namespace`/`queue` properties, but you are right that `is_ready` and `relation_payloads` still duplicate parts of the same validity check. I have a small follow-up refactor prepared to centralize this in one helper and can push it if you want this tightened before merge.

terraform/justfile
nikita-jare: Good question. This retry was added after a Terraform checks failure where `rockcraft pack` failed due transient apt mirror sync issues inside the build instance (`Failed to fetch ... File has unexpected size ... Mirror sync in progress`). The failing run was https://github.com/canonical/temporal-worker-k8s-operator/actions/runs/24519879737 and this happened in the `Terraform apply (test)` job. The same change also tightened artifact handling (expect exactly one `.rock`) so we fail deterministically instead of later passing a bad path to skopeo.

PR #87, opened 2025-12-05
[WF-203] Use juju provider v1.0.0

jananisenthilkumar291:
## Description

This PR updates the Terraform integration for the temporal-ui-k8s-operator to use the new Juju provider v1.0.0, replaces the deprecated model attribute with model_uuid, and removes the need for specifying a model or UUID in the module inputs.

The model UUID is now resolved using the provider’s juju_model data source, ensuring that Terraform always deploys into the correct Juju model without requiring manual configuration or Justfile-injected values.

## Key-Changes:
- Replace model attribute with `model_uuid`. 
-  New reuired field is `model_uuid = data.juju_model.terraform.uuid` 
- Introduce a `juju_model` data source to resolve model_uuid dynamically.
- Remove obsolete `model` variable declaration from `variables.tf`
- Upgrade provider version to v1.0.0
- `test/test.tfvars` no longer needs to declare `model = "terraform"`
- Removed custom local-exec logic for attaching oci-images and replaced that with juju provider-native resources and registry_credentials fields. 
- This change allows juju provider to manage image resources directly.
---
## Testing instructions

just validate
just fmt
just lint
just test

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/terraform/justfile
+++ b/terraform/justfile
@@ -38,7 +38,10 @@ destroy-model:
 
 [private]
 add-model: (destroy-model)
+    #!/usr/bin/bash
     juju add-model terraform
+    MODEL_UUID=$(juju show-model terraform --format=json | jq -r '.terraform["model-uuid"]')
+    sed -i "s|model_uuid = \".*\"|model_uuid = \"${MODEL_UUID}\"|" test/test.tfvars
 
 # Run terraform format
 fmt: (initialize)
--- a/terraform/justfile
+++ b/terraform/justfile
@@ -67,9 +67,11 @@ apply variables_file: (initialize)
 
 # Destroy the terraform plan with the provided variables file
 destroy variables_file:
+    #!/usr/bin/bash
     terraform destroy -auto-approve -var-file="${variables_file}"
 
     just destroy-model
+    sed '/model_uuid/d' -i "${variables_file}"
 
 test: (build-and-push-image)
     #!/usr/bin/bash
--- a/terraform/justfile
+++ b/terraform/justfile
@@ -38,7 +38,20 @@ destroy-model:
 
 [private]
 add-model: (destroy-model)
+    #!/usr/bin/bash
     juju add-model terraform
+
+    MODEL_UUID=$(juju show-model terraform --format=json | jq -r '.terraform["model-uuid"]')
+    echo "model_uuid = \"${MODEL_UUID}\"" >> "./test/test.tfvars"
+
+[private]
+validate_test_tfvars variables_file="test/test.tfvars":
+    #!/usr/bin/bash
+    MODEL_UUID=$(grep '^model_uuid' "${variables_file}" | awk -F '"' '{print $2}')
+
+    if ! juju show-model "${MODEL_UUID}" >/dev/null 2>&1; then
+        exit 1
+    fi
 
 # Run terraform format
 fmt: (initialize)
@@ -70,6 +83,8 @@ test: (build-and-push-image)
 
     just add-model
 
+    just validate_test_tfvars "./test/test.tfvars"
+    
     just apply "./test/test.tfvars"
 
     juju wait-for model terraform --query="forEach(units, unit => (unit.workload-status == 'blocked' && unit.workload-message == 'Invalid config: host value missing'))"
--- a/terraform/justfile
+++ b/terraform/justfile
@@ -40,8 +40,18 @@ destroy-model:
 add-model: (destroy-model)
     #!/usr/bin/bash
     juju add-model terraform
+
     MODEL_UUID=$(juju show-model terraform --format=json | jq -r '.terraform["model-uuid"]')
     sed -i "s|model_uuid = \".*\"|model_uuid = \"${MODEL_UUID}\"|" test/test.tfvars
+
+[private]
+validate_test_tfvars variables_file="test/test.tfvars":
+    #!/usr/bin/bash
+    MODEL_UUID=$(grep '^model_uuid' "${variables_file}" | awk -F '"' '{print $2}')
+    
+    if ! juju show-model "${MODEL_UUID}" >/dev/null 2>&1; then
+        exit 1
+    fi
 
 # Run terraform format
 fmt: (initialize)
@@ -72,7 +82,7 @@ test: (build-and-push-image)
     trap 'just destroy "./test/test.tfvars"' EXIT
 
     just add-model
-
+    just validate_test_tfvars "./test/test.tfvars"
     just apply "./test/test.tfvars"
 
     juju wait-for model terraform --query="forEach(units, unit => (unit.workload-status == 'blocked' && unit.workload-message == 'Invalid config: host value missing'))"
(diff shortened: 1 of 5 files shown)


terraform/justfile
shayancanonical, line 83: before `just apply`, i'd add commands to assert that the model_uuid in `test.tfvars` is a valid juju model. additionally, this can be a just target (`validate_test_tfvars`)

terraform/main.tf
```diff
@@ -11,30 +11,16 @@ resource "juju_application" "temporal_worker_k8s" {
...
-  provisioner "local-exec" {
-    # Needed since juju_application resource does not support resource map to specify registry creds
-    # Refactor once https://github.com/juju/terraform-provider-juju/issues/620 resolved
-    command = <<EOT
-      echo "${yamlencode({
-    "registrypath" = var.image.image,
-    "username"     = var.image.registry_username,
-    "password"     = var.image.registry_password,
-})}" > temporal_worker_image.yaml
+  #  registry_credentials = {
```
shayancanonical: should this be commented? we should be specifying the image repo credentials if present in input vars

terraform/variables.tf
```diff
@@ -36,10 +36,14 @@ variable "constraints" {
 variable "image" {
   type = object({
     image             = string
+    image_repository  = optional(string, "")
     registry_username = optional(string, "")
     registry_password = optional(string, "")
   })
   description = "Details of the worker image"
+  default = {
```
shayancanonical: why specify a default at all? if default is not specified, terraform will enforce the variable as required, in which case, the check for `var.image.image != ""` can be dropped

jananisenthilkumar291: Initially I assumed a default value was required to avoid the `unknown-value` issue, because using a default guarantees that Terraform always has a concrete value for `image.image`. However, the real issue turned out not to be whether the variable is required, but rather how Terraform evaluates object variables containing optional attributes.

Even with a required variable and no default, terraform still seems to treat the object as partially unknown during the `plan` phase, until it has resolved all the object's _required_ attributes (image.image) provided by user (in this case test/test.tfvars). Therefore, the conditional is required regardless of whether the variable has a default.

I'm making the change to remove `default{}` for `image.image`.

terraform/main.tf
jananisenthilkumar291: Good catch! It was a miss on my side. I'm updating that in the next commit.

terraform/justfile
jananisenthilkumar291: Makes Sense. Adding the just target in next push.

terraform/test/test.tfvars
```diff
@@ -1,3 +1,5 @@
-model   = "terraform"
-image   = { "image" : "docker://localhost:5000/temporal-worker:test-terraform" }
+model_uuid = "00000000-0000-0000-0000-000000000000"
```
shayancanonical: i would recommend removing this. if the model_uuid is not included in the tfvars file, terraform should either complain or the juju provider should figure out how to deploy to current selected model. i prefer not having any invalid contents in this file as we have to avoid assumptions that people will exclusively use `just` targets to apply

please adjust downstream changes accordingly

jananisenthilkumar291: Noted! That seems much cleaner, instead of injecting model_uuid.
I removed `model_uuid` from `test/test.tfvars` and replaced it with a data `"juju_model"` lookup. This allows the Juju provider to infer the correct model UUID automatically, eliminates invalid test data, removes Justfile-dependent behavior, and ensures Terraform behaves correctly whether the model exists or not.

terraform/main.tf
```diff
@@ -1,6 +1,11 @@
+data "juju_model" "terraform" {
+  name  = "terraform"
+  owner = "admin"
```
shayancanonical: not sure that we want to hard code the name of the model to `terraform`

note: when we run the `terraform apply` as tests in this repo, the model name is `terraform`. when the module is used by our users, they will have a `resource "juju_model" "model_name" {}` section in their tf files. the spec that guides how we sync our tf modules indicates that `model_uuid` should be an input var

recommend taking in `model_uuid` (a required input) in the module. additionally, we should omit `model_uuid` from the `test.tfvars` file. if someone applies the `test.tfvars` with the module, then at some point, terraform will throw an error that the model_uuid is not found (may be not right away, but when the juju provider realizes this)

shayancanonical (changes requested):

terraform/main.tf
jananisenthilkumar291: Agreed! It's efficient to not have model name hard coded.

On another note, 
If we want to omit model_uuid from test.tfvars, for the sake of `just apply` succeeding, we'll need to inject the model_uuid dynamically (Something we tried [here](https://github.com/canonical/temporal-worker-k8s-operator/pull/87/commits/ae571ee7161df74e312e6fabf2ab46cb45f44b71)).
A way we can achieve [juju provider figuring out model_uuid](https://github.com/canonical/temporal-worker-k8s-operator/pull/87#discussion_r2593549087) is by having a data source which helps terraform juju provider to fetch model_uuid when it's provided with model name as per the documentaion [here](https://documentation.ubuntu.com/terraform-provider-juju/latest/tutorial/#:~:text=resource%20%22juju_model%22%20%22chat%22%20%7B%0A%20%20name%20%3D%20%22chat%22%0A%7D%0A%0Aresource%20%22juju_application%22%20%22mattermost%2Dk8s%22%20%7B%0A%20%20model_uuid%20%3D%20juju_model.chat.uuid).
Having model name as an input just makes it a lot simpler rather than having the user lookup for model_uuid after creating the model and then do a `terraform apply`. juju provider handles fetching model_uuid on it's own.

jananisenthilkumar291: Going strictly by spec requirements the [commit](https://github.com/canonical/temporal-worker-k8s-operator/pull/87/commits/e5c6ce88c22cd7408d83c206586c4149b1c6402a) should make sense. This ensures, model_uuid as a mandatory variable input from user. It's injected using out just target into `test.tfvars`.
a regular `terraform apply` will fail if users do not add model_uuid.

terraform/main.tf
```diff
@@ -11,30 +11,16 @@ resource "juju_application" "temporal_worker_k8s" {
...
+  resources = var.image.image != "" ? {
+    "temporal-worker-image" = var.image.image
+  } : {}
 
-      rm temporal_worker_image.yaml
-    EOT
-}
-
-depends_on = [resource.juju_application.temporal_worker_k8s]
+  registry_credentials = {
```
shayancanonical: would we need a condition here as well? if the `image_repository` is empty, do we want to include `registry_credentials`?

jananisenthilkumar291: Fair point! Thanks for the catch Shayan. 
We do not want to set a registry credential for an empty repository.
The [commit](https://github.com/canonical/temporal-worker-k8s-operator/pull/87/commits/b584857afbeee91deeba7b0a100f572964126d9d) addresses the following
- if repo is provided, valid credentials map is set
- else, empty credentials map is set.

shayancanonical (approved):
lgtm!

PR #91, opened 2026-03-24
[WF-144] Fix env precedence over charm config

### LINKED ISSUE #36: Cannot override temporal config files using environmental variables.

nikita-jare:
## Description

Fixes config precedence so values from environment (env/juju/vault) are no longer overwritten by charm config defaults.

## Closes #36

---

- `_update` now applies self.config-derived TWC_/TEMPORAL_ values first, then overlays `create_env()` values from the environment config.
- Added scenario tests to cover:
   env/juju/vault precedence
   Juju secret override of charm config
   auth-secret-id staying highest priority for auth keys
---
## Checklist (all that apply)
- [X] Unit tests added or updated
- [ ] Documentation updated

### DIFF
--- a/tests/scenario/test_charm.py
+++ b/tests/scenario/test_charm.py
@@ -144,6 +144,39 @@ WANT_ENV_AUTH = {
 }
 
 
+def _build_environment_config(secret_id: str, include_env: bool = False, include_vault: bool = False) -> str:
+    """Build environment config YAML for encryption-key precedence tests."""
+    env_section = (
+        """
+        env:
+            - name: TEMPORAL_ENCRYPTION_KEY
+              value: env-value
+        """
+        if include_env
+        else ""
+    )
+    vault_section = (
+        """
+        vault:
+            - path: secrets
+              name: TEMPORAL_ENCRYPTION_KEY
+              key: token
+        """
+        if include_vault
+        else ""
+    )
+    return textwrap.dedent(
+        f"""
+        {env_section}
+        juju:
+            - secret-id: {secret_id}
+              name: TEMPORAL_ENCRYPTION_KEY
+              key: encryption-key
+        {vault_section}
+        """
+    )
+
+
 @pytest.fixture
 def all_required_relations(peer_relation, vault_relation, database_relation):
     return [
@@ -436,29 +469,45 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
         )
 
 
-def test_environment_juju_secret_overrides_empty_charm_config(
-    context, state, temporal_worker_container, config, encryption_key_secret, vault_nonce_secret
+@pytest.mark.parametrize(
+    "charm_encryption_key, expected_temporal_key, expected_twc_key",
+    [
+        ("", "secret-encryption-key", ""),
+        ("plaintext-config-key", "secret-encryption-key", "plaintext-config-key"),
+    ],
+    ids=["empty-charm-config", "non-empty-charm-config"],
+)
+def test_environment_juju_secret_overrides_charm_config(
+    context,
+    state,
+    temporal_worker_container,
+    config,
+    encryption_key_secret,
+    vault_nonce_secret,
+    charm_encryption_key,
+    expected_temporal_key,
+    expected_twc_key,
 ):
     state = dataclasses.replace(state, secrets=[encryption_key_secret, vault_nonce_secret])
     state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
     state_out = context.run(context.on.config_changed(), state_out)
 
-    environment_config = textwrap.dedent(
-        f"""
-        juju:
-            - secret-id: {encryption_key_secret.id}
-              name: TEMPORAL_ENCRYPTION_KEY
-              key: encryption-key
-        """
-    )
+    environment_config = _build_environment_config(secret_id=encryption_key_secret.id)
 
-    state_out = dataclasses.replace(state_out, config={**config, "environment": environment_config})
+    state_out = dataclasses.replace(
+        state_out,
+        config={**config, "encryption-key": charm_encryption_key, "environment": environment_config},
+    )
     with unittest.mock.patch(
         "ops.jujuversion.JujuVersion.from_environ", return_value=ops.jujuversion.JujuVersion(version="3.6")
     ):
         state_out = context.run(context.on.config_changed(), state_out)
 
-    expected_env = {**WANT_ENV, "TEMPORAL_ENCRYPTION_KEY": "secret-encryption-key"}
+    expected_env = {
+        **WANT_ENV,
+        "TEMPORAL_ENCRYPTION_KEY": expected_temporal_key,
+        "TWC_ENCRYPTION_KEY": expected_twc_key,
+    }
     assert sorted(state_out.get_container("temporal-worker").plan.to_dict()) == sorted(
         {
             "services": {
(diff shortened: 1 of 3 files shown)


tests/scenario/test_charm.py
```diff
@@ -436,6 +436,187 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
         )
 
 
+def test_environment_juju_secret_overrides_empty_charm_config(
```
jananisenthilkumar291: I'd prefer using `pytest.mark.parameterize` to test different test cases (in a single test definition) rather than having different tests written for each case.

tests/scenario/test_charm.py
```diff
@@ -436,6 +436,187 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+
+
+def test_environment_precedence_vault_over_juju_over_env(
+    context, state, temporal_worker_container, config, encryption_key_secret, vault_nonce_secret
+):
+    state = dataclasses.replace(state, secrets=[encryption_key_secret, vault_nonce_secret])
+    state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
+    state_out = context.run(context.on.config_changed(), state_out)
+
+    environment_config = textwrap.dedent(
```
jananisenthilkumar291: Suggestion: 
Prefer having a helper to avoid repeating the textwrap logic.

jananisenthilkumar291 (commented):
Couple suggestions in tests. Rest LGTM. Thanks, @nikita-jare

tests/scenario/test_charm.py
nikita-jare: Refactored in the latest commit [0664d8d](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/commits/0664d8d8bc66410c3ef7a40c612b362aef5187ee)

tests/scenario/test_charm.py
nikita-jare: Updated in the latest commit [0664d8d](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/commits/0664d8d8bc66410c3ef7a40c612b362aef5187ee)

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+    ), unittest.mock.patch(
+        "builtins.open", new_callable=unittest.mock.mock_open
+    ):
+        mock_vault_client = unittest.mock.Mock()
+        mock_vault_client.read_secret.return_value = "vault-value"
+        get_vault_client.return_value = mock_vault_client
+        state_out = context.run(context.on.config_changed(), state_out)
+
+    expected_env = {**WANT_ENV, "TEMPORAL_ENCRYPTION_KEY": "vault-value"}
+    assert sorted(state_out.get_container("temporal-worker").plan.to_dict()) == sorted(
```
jananisenthilkumar291: If we intend to test only the `expected_env`, why not just assert that, rather than the entire dict.
```suggestion
actual_env = state_out.get_container("temporal-worker").plan.to_dict()["services"]["temporal-worker"]["environment"]
    assert actual_env == {**WANT_ENV, "TEMPORAL_ENCRYPTION_KEY": "vault-value"}
```

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+        }
+    )
+
+
+def test_environment_precedence_vault_over_juju_over_env(
+    context, state, temporal_worker_container, config, encryption_key_secret, vault_nonce_secret
+):
+    state = dataclasses.replace(state, secrets=[encryption_key_secret, vault_nonce_secret])
+    state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
+    state_out = context.run(context.on.config_changed(), state_out)
```
jananisenthilkumar291: I think, this is redundant here.
Since we are changing the config anyway, we might only need to run the hook once at the end(which is done [here](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/changes#diff-ad400e12619d537dbe2302a9f251700e412661acd792494861e9f8f633904220R552)) to verify the resulting Pebble plan. But please confirm.

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+        }
+    )
+
+
+def test_auth_secret_overrides_environment_auth_keys(
+    context, state, temporal_worker_container, config, encryption_key_secret, oidc_auth_secret, vault_nonce_secret
+):
+    state = dataclasses.replace(state, secrets=[encryption_key_secret, oidc_auth_secret, vault_nonce_secret])
+    state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
+    state_out = context.run(context.on.config_changed(), state_out)
```
jananisenthilkumar291: Similar to the above test, 
we don't need to run every intermediate hook if the final hook (the second config_changed) sets the state exactly as we need it.[here](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/changes#diff-ad400e12619d537dbe2302a9f251700e412661acd792494861e9f8f633904220R586)

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+                "temporal-worker": {
+                    "summary": "temporal worker",
+                    "command": "/app/scripts/start-worker.sh",
+                    "startup": "enabled",
+                    "override": "replace",
+                    "environment": expected_env,
+                },
+            },
+        }
+    )
```
jananisenthilkumar291: We could collapse this to assert only the key we intend to test. 
```suggestion
    actual_env = state_out.get_container("temporal-worker").plan.to_dict()["services"]["temporal-worker"]["environment"]
    assert actual_env == {**WANT_ENV, **WANT_ENV_AUTH}
```

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+    config,
+    encryption_key_secret,
+    vault_nonce_secret,
+    charm_encryption_key,
+    expected_temporal_key,
+    expected_twc_key,
+):
+    state = dataclasses.replace(state, secrets=[encryption_key_secret, vault_nonce_secret])
+    state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
+    state_out = context.run(context.on.config_changed(), state_out)
```
jananisenthilkumar291: Please confirm if this is absolutely necessary, as we're running the hook again [here](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/changes#diff-ad400e12619d537dbe2302a9f251700e412661acd792494861e9f8f633904220R504), finally after the config is changed.

tests/scenario/test_charm.py
```diff
@@ -436,6 +469,139 @@ def test_valid_environment_config(context, state, temporal_worker_container, con
...
+        "ops.jujuversion.JujuVersion.from_environ", return_value=ops.jujuversion.JujuVersion(version="3.6")
+    ):
+        state_out = context.run(context.on.config_changed(), state_out)
+
+    expected_env = {
+        **WANT_ENV,
+        "TEMPORAL_ENCRYPTION_KEY": expected_temporal_key,
+        "TWC_ENCRYPTION_KEY": expected_twc_key,
+    }
+    assert sorted(state_out.get_container("temporal-worker").plan.to_dict()) == sorted(
```
jananisenthilkumar291: Similar suggestion as [here](https://github.com/canonical/temporal-worker-k8s-operator/pull/91/changes#r3016555638).

jananisenthilkumar291 (commented):
Added non-blocking comments. Rest LGTM!

tests/scenario/test_charm.py
nikita-jare: removed the intermediate config_changed

tests/scenario/test_charm.py
nikita-jare: dropped the full plan.to_dict() comparison and assert the environment block

jananisenthilkumar291 (dismissed):
LGTM! Thanks @nikita-jare

shayancanonical (dismissed):
lgtm

nit suggestion: add comments to introduced tests (add summary so it is much easier to trace and quickly understand the assertions/validations of the tests)

shayancanonical (approved):

PR #94, opened 2026-04-29
[WF-404] Migrate to Ubuntu 24.04

jananisenthilkumar291:
## Description
Migrate the Temporal worker charm to Ubuntu 24.04.

Changes:
- `charmcraft.yaml`: Updated platform from `ubuntu@22.04:amd64` to `ubuntu@24.04:amd64`
- `tests/integration/test_upgrades.py`: Replaced `application.refresh()` (Python libjuju) with `ops_test.juju()` CLI call and added `--base ubuntu@24.04` parameter. Python libjuju doesn't support the `--base` parameter required for cross-base upgrades due to Juju bug [#20031](https://github.com/juju/juju/issues/20031)
- `tox.ini`: Modified bandit command to scan only `src/` directory (excluded `tests/`). Test fixtures contain dummy passwords that bandit 1.9.3 flags as false-positive security issues

---
## Testing instructions

1. Build charm with `charmcraft pack`
2. Verify the built charm targets `ubuntu@24.04`
3. Run integration upgrade tests to verify cross-base refresh works correctly

The upgrade test now uses `ops_test.juju("refresh", ..., "--base", "ubuntu@24.04")` instead of libjuju's `application.refresh()`. This is necessary because libjuju doesn't expose the `--base` parameter, which Juju requires when upgrading across base versions.

---
## Checklist (all that apply)
- [x] Integration tests added or updated
- [ ] Documentation updated

### DIFF
--- a/charmcraft.yaml
+++ b/charmcraft.yaml
@@ -4,7 +4,7 @@
 type: charm
 
 platforms:
-  ubuntu@22.04:amd64:
+  ubuntu@24.04:amd64:
 
 parts:
   charm:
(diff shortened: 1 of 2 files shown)


nikita-jare (approved):
Thank you @jananisenthilkumar291

PR #95, opened 2026-04-30
fix: remove cancel-in-fail from integration tests workflow

DnPlas:
## Description

When a workflow is called via workflow_call, the concurrency should be controlled by the parent workflow, not the child. This commit removes the concurrency field from the integration tests workflow to prevent issues when reused by another workflow.

---
## Testing instructions

N/A

This commit resolves the issues observed [here](https://github.com/canonical/temporal-worker-k8s-operator/actions/runs/25166155149/job/73841609682).

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/.github/workflows/integration_test.yaml
+++ b/.github/workflows/integration_test.yaml
@@ -1,8 +1,5 @@
 name: Integration tests
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
 on:
   pull_request:
   workflow_call:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,5 +1,9 @@
 name: CI
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: false
+
 on:
   push:
     branches:


.github/workflows/integration_test.yaml
```diff
@@ -1,8 +1,5 @@
 name: Integration tests
 
-concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
```
nikita-jare, line 5: Suggestion: since the concurrency control should be handled by the parent, should we add the concurrency block to ci.yaml so push-triggered runs on main/track/** still get cancel-on-supersede behavior?

nikita-jare (commented):
LGTM! Thank you @DnPlas 
Just one suggestion/question below.

.github/workflows/integration_test.yaml
DnPlas: I have added the concurrency in the ci.yaml, BUT I am not cancelling the job, instead I am queueing them so they do not get cancelled mid-way. Resolved in [044c19b](https://github.com/canonical/temporal-worker-k8s-operator/commit/044c19b52290fc3ffe0017e53ba64368f02a63ae)

nikita-jare (approved):
Thank you @DnPlas!

PR #103, opened 2026-05-26
fix: pin kubernetes<36 to avoid v36.0.0 auth regression

nikita-jare:
## Description

kubernetes v36.0.0 introduced a regression where load_incluster_config() silently drops the Authorization header, causing all in-cluster API calls to fail with 401. See kubernetes-client/python#2584.

Pin kubernetes<36 to prevent pip from resolving to the broken version at install time. Remove the upper bound once a patched release is available.

---

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -32,6 +32,8 @@ unit = [
 integration = [
     "ipdb==0.13.9",
     "juju==3.5.2.1",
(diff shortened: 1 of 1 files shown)


shayancanonical (dismissed):
lgtm, i would also add an inline comment to remove pin once issue is solved (with link to GH issue)

shayancanonical (approved):

PR #105, opened 2026-05-27
[WF-190] fix: detect service crash/restart loops in update-status

### LINKED ISSUE #104: Crashes/Restart Loops are not detected
## Summary

When the `temporal-worker` container service crashes and enters a Pebble restart loop (on-failure: restart), the Juju unit status remains **green / Active / Idle**. Operators have no indication that the workload is repeatedly crashing and is never actually healthy.

---

## Observed Behaviour

Pebble logs showed the service exiting with code 1 and being restarted every ~30s (exponential backoff, counter in the hundreds), see below logs.

Despite this, `juju status` reported the application and unit as:

```
App                      Version  Status  Scale  Charm                    Channel  Rev  ...
temporal-worker-k8s               active      1  temporal-worker-k8s      ...      ...

Unit                        Workload  Agent  ...  Message
temporal-worker-k8s/0*      active    idle   ...  worker listening to namespace '...' on queue '...'
```

---

## Root Cause Analysis

The `_on_update_status` handler in `src/charm.py` validates the Pebble *plan* (i.e. whether the YAML service definition exists) but does **not** check whether the service process is actually **running**.

---

## Impact

- Operators see a **false green** status for a unit whose workload is not serving any requests.
- Silent restart loops can last indefinitely (Pebble's backoff counter was at 444+ in the incident above), with no Juju issue surfaced.
- Root-cause investigation is delayed because the incident is only discoverable by tailing raw Pebble/pods logs.

---

## Expected Behaviour

When `update-status` fires and the `temporal-worker` service is not running, the unit status should change to **Blocked** with a clear message directing operators to the logs:

```
Unit                        Workload  Agent  Message
temporal-worker-k8s/0*      blocked   idle   temporal-worker service is not running; check logs for crash details
```

### To Reproduce

1. Deploy `temporal-worker-k8s-operator` with a worker image whose application raises an unhandled exception on startup (e.g. duplicate activity registration: `ValueError: More than one activity named <name>`).
3. Observe that the unit shows **active/idle** despite the worker never running successfully.

### Environment

Juju controller: 3.5.5

### Relevant log output

```shell
2026-05-25T06:52:16.418Z [pebble] Service "temporal-worker" stopped unexpectedly with code 1
2026-05-25T06:52:16.418Z [pebble] Service "temporal-worker" on-failure action is "restart", waiting ~30s before restart (backoff 444)
2026-05-25T06:52:48.423Z [pebble] Service "temporal-worker" starting: ./app/scripts/start-worker.sh
2026-05-25T06:52:53.177Z [temporal-worker] ValueError: More than one activity named deactivate_user_by_id
2026-05-25T06:52:53.523Z [pebble] Service "temporal-worker" stopped unexpectedly with code 1
2026-05-25T06:52:53.523Z [pebble] Service "temporal-worker" on-failure action is "restart", waiting ~30s before restart (backoff 445)
```

### Additional context

I'll push a PR for fixing this bug shortly.

kot0dama:
## Description

The charm was falsely reporting Active status even when the temporal-worker service was caught in a Pebble crash/restart loop (on-failure: restart).

This fixes #104.

Root cause: _on_update_status only validated the existence of the Pebble plan (via _validate_pebble_plan) but never checked whether the service process was actually running.  When the workload exits with a non-zero code (e.g. ValueError: More than one activity named <x>) and Pebble is in backoff between restart attempts, get_service() returns ServiceStatus.INACTIVE.  The charm never called get_service(), so it unconditionally set ActiveStatus after a successful plan check and never detected the crashed state.

```python
# Before fix — only checks the plan, never the live process state
valid_pebble_plan = self._validate_pebble_plan(container)
if not valid_pebble_plan:
    self._update(event)
    return

# Reached even when the service is INACTIVE / in restart backoff
self.unit.status = ActiveStatus(...)
```

Fix: in _on_update_status, call container.get_service() after the plan validation.  If the service is not running, set BlockedStatus with a message directing operators to the logs, and return early.

Observed failure log:

>   [pebble] Service "temporal-worker" stopped unexpectedly with code 1
>   [pebble] Service "temporal-worker" on-failure action is "restart",
>            waiting ~30s before restart (backoff 444)
>   [temporal-worker] ValueError: More than one activity named deactivate_user_by_id

Also adds test_service_crash_restart_loop_detected scenario test

---
## Checklist (all that apply)
- [x] Unit tests added or updated
- [ ] Documentation updated

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -66,6 +66,8 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
         self.framework.observe(self.on.restart_action, self._on_restart)
         self.framework.observe(self.on.update_status, self._on_update_status)
         self.framework.observe(self.on.install, self._on_install)
+        self.framework.observe(self.on.temporal_worker_pebble_check_failed, self._on_pebble_check_failed)
+        self.framework.observe(self.on.temporal_worker_pebble_check_recovered, self._on_pebble_check_recovered)
         self.framework.observe(self.on.secret_changed, self._on_secret_changed)
 
         # Vault
@@ -195,6 +197,37 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             self._update(event)
             return
 
+        try:
+            service = container.get_service(self.name)
+            if not service.is_running():
+                logger.error(
+                    "temporal-worker service is not running; it may be caught in a crash/restart loop - check logs for details"
+                )
+                self.unit.status = BlockedStatus("temporal-worker service is not running; check logs for crash details")
+                return
+        except pebble.APIError as e:
+            logger.warning(f"Could not retrieve service status: {e}")
+
+        self.unit.status = ActiveStatus(
+            f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
+        )
+
+    @log_event_handler(logger)
+    def _on_pebble_check_failed(self, event):
+        """Handle pebble check failed event.
+
+        Args:
+            event: The event triggered when the pebble check fails.
+        """
+        self.unit.status = BlockedStatus("temporal-worker service is not running; check logs for crash details")
+
+    @log_event_handler(logger)
+    def _on_pebble_check_recovered(self, event):
+        """Handle pebble check recovered event.
+
+        Args:
+            event: The event triggered when the pebble check recovers.
+        """
         self.unit.status = ActiveStatus(
             f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
         )
@@ -445,6 +478,13 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
                     "startup": "enabled",
                     "override": "replace",
                     "environment": context,
+                }
+            },
+            "checks": {
+                "start-worker-check": {
+                    "override": "replace",
+                    "threshold": 3,
+                    "exec": {"command": "pgrep -f start-worker.sh"},
                 }
             },
         }
--- a/src/charm.py
+++ b/src/charm.py
@@ -197,20 +197,6 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             self._update(event)
             return
 
-        try:
-            service = container.get_service(self.name)
-            if not service.is_running():
-                logger.error(
-                    "temporal-worker service is not running; it may be caught in a crash/restart loop - check logs for details"
-                )
-                self.unit.status = BlockedStatus("temporal-worker service is not running; check logs for crash details")
-                return
-        except pebble.APIError as e:
-            logger.warning(f"Could not retrieve service status: {e}")
-
-        self.unit.status = ActiveStatus(
-            f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
-        )
 
     @log_event_handler(logger)
     def _on_pebble_check_failed(self, event):
@@ -500,7 +486,9 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             )
             return
 
-        self.unit.status = MaintenanceStatus("replanning application")
+        self.unit.status = ActiveStatus(
+            f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
+        )
 
 
 def convert_env_var(config_var, prefix="TWC_"):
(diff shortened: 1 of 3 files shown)


src/charm.py
```diff
@@ -195,6 +195,19 @@ def _on_update_status(self, event):
...
+            service = container.get_service(self.name)
+            if not service.is_running():
+                logger.error(
+                    "temporal-worker service is not running; it may be caught in a crash/restart loop - check logs for details"
+                )
+                self.unit.status = BlockedStatus(
+                    "temporal-worker service is not running; check logs for crash details"
+                )
+                return
+        except pebble.APIError as e:
```
nikita-jare: `_validate_pebble_plan` a few lines above catches `pebble.ConnectionError`. If pebble drops the connection between the plan check and `get_service`, this will raise instead of being handled gracefully. Suggest catching `pebble.ConnectionError` too, same pattern as the helper above.

src/charm.py
```diff
@@ -195,6 +195,19 @@ def _on_update_status(self, event):
...
+            logger.warning(f"Could not retrieve service status: {e}")
```
nikita-jare: Could you also return inside the except instead of just logging? If `get_service` raises, we currently fall through to ActiveStatus below, which is the same shape as the bug this PR is fixing. Returning here keeps the behavior consistent with `_validate_pebble_plan`.

nikita-jare (commented):
thanks @kot0dama for the detailed root-cause analysis in the issue. Approach looks good overall.

src/charm.py
kot0dama: Thank you! After a chat with @DnPlas I've made some other changes to use Pebble notices, not sure if this is still relevant, feel free to resolve or reply!

src/charm.py
kot0dama: Thank you! After a chat with @DnPlas I've made some other changes to use Pebble notices, not sure if this is still relevant, feel free to resolve or reply!

kot0dama:
@DnPlas I can't seem to be able to shake that CI failure in the new `test_service_crash_and_recovery` test.
I had this, but removed now:
```diff
+def test_service_crash_and_recovery(context, state, temporal_worker_container, namespace, queue):
+    state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
+    state_out = context.run(context.on.config_changed(), state_out)
+
+    # We must retrieve the container from the updated state so that
+    # scenario knows the 'start-worker-check' actually exists in the plan.
+    updated_container = state_out.get_container("temporal-worker")
+
+    # Create the CheckInfo object that ops-scenario expects, using the
+    # check name defined in the charm.
+    check_info = ops.pebble.CheckInfo(
+        name="start-worker-check", level=ops.pebble.CheckLevel.ALIVE, status=ops.pebble.CheckStatus.UP
+    )
+
+    # ops-scenario has a consistency checker bug with container names containing dashes.
+    # It resolves 'temporal_worker_pebble_check_failed' to 'temporal_worker' instead of 'temporal-worker'.
+    # We bypass the consistency check to avoid this false-positive error.
+    with unittest.mock.patch("scenario._runtime.check_consistency"):
+        # Simulate the pebble check failing
+        state_out = context.run(context.on.pebble_check_failed(updated_container, check_info), state_out)
+
+        assert state_out.unit_status == ops.BlockedStatus(
+            "temporal-worker service is not running; check logs for crash details"
+        )
+
+        # Simulate the pebble check recovering
+        state_out = context.run(context.on.pebble_check_recovered(updated_container, check_info), state_out)
+
+        assert state_out.unit_status == ops.ActiveStatus(
+            f"worker listening to namespace {namespace!r} on queue {queue!r}"
+        )
+
+
```

Feel free to merge this one without the additional test and add a similar unit test !

DnPlas:
I have updated the PR slightly to avoid checking the status of the service in the update_status event handler. To also avoid overriding the status of the other status handlers, I have removed the `ActiveStatus` set in the update_status and instead I am only setting it after all of the checks in `_update()` run successfully.

Thanks @kot0dama for your contributions, we'll take it from here. Requesting the @canonical/workflows team to review this PR.

nikita-jare (approved):
The pivot to a Pebble check is a better design than the earlier polling approach. Looks good to me. Thank you @kot0dama and @DnPlas

shayancanonical (approved):
lgtm

kot0dama:
Thank you @DnPlas @nikita-jare @shayancanonical for the reviews and guidance!
Glad we were able to fix this issue promptly :slightly_smiling_face:

PR #107, opened 2026-05-27
fix: detect Temporal SDK workflow eviction loop in update-status

### LINKED ISSUE #106: Workflow Eviction Loop Not Surfaced in Charm Status
### Bug Description

When the Temporal Python SDK enters a workflow eviction loop (due to a blocking call inside a workflow coroutine), the Juju unit remains Active / Idle and does not surface the error.

The worker keeps running (Pebble service status: ACTIVE), so neither the crash detection fix nor a simple service-status check catches it.  Meanwhile, affected workflows are stuck indefinitely cycling between `WORKFLOW_TASK_FAILED` and `WORKFLOW_TASK_TIMED_OUT`.

## What Is a Workflow Eviction Loop?

Temporal caches active workflow instances in the worker's in-process cache (the "sticky task queue" cache).  When a workflow task fails, Temporal attempts to evict the cached instance so it can replay history on a fresh instance.

If the eviction itself fails (for example because there is still an asyncio task running that cannot be cancelled), the SDK logs:

```
ERROR:temporalio.worker._workflow:Failed running eviction job for run ID <X>,
continually retrying eviction. Since eviction could not be processed, this worker
may not complete and the slot may remain forever used unless it eventually completes.
```

The worker process does **not** exit.  Pebble reports the service as ACTIVE. The slot is occupied indefinitely, and Temporal keeps scheduling the workflow task back to the same (stuck) worker, where it fails/times out again.

---

### In `hr-automation` (primary trigger)

1. `SalesloftLeaverWorkflow` child workflow exhausted 10 retries with a proxy error: `ProxyError: Tunnel connection failed: 403 Forbidden`
2. `asyncio.gather(return_exceptions=True)` returned in the parent workflow `HRDelegatedOffboardingWorkflow`.
3. `process_plugin_result()` called `workflow_logger.info(...)` with a `LogExtra` `extra=` dict.
4. `workflow_logger.base_logger` had `PostgresHandler` **directly attached** (set up in `configure_temporal_logger`).  `PostgresHandler.emit()` performs a **synchronous blocking SQLAlchemy INSERT** into the `audit_logs` table.
5. The blocking DB write held the workflow coroutine for > 2 seconds, triggering Temporal's deadlock detector:
   ```
   [TMPRL1101] Potential deadlock detected: workflow didn't yield within 2 second(s).
   ```
6. Temporal failed the workflow task.  On the next replay, it tried to evict the cached workflow instance, but the asyncio task was still pending in the coroutine runner, eviction failed.
7. The SDK entered the eviction retry loop, logging the sentinel error above.
8. The worker continued running normally from Pebble's perspective.

### In the charm (detection gap)

`_on_update_status` only checked:
1. Whether the Pebble plan was valid (✓)
2. Whether the service was running (✓ ACTIVE)

It did **not** scan the service logs for the Temporal SDK eviction error, so the unit was reported as `active/idle` even though it could not process the affected workflow tasks.

---

## Observed Pebble Log Sequence

See below.

## Temporal Workflow History (f04e62c4-...)

| Event | Type |
|-------|------|
| 28 | `CHILD_WORKFLOW_EXECUTION_FAILED` (SalesloftLeaverWorkflow) |
| 29–30 | `WORKFLOW_TASK_SCHEDULED / STARTED` |
| 31 | `WORKFLOW_TASK_FAILED` (_DeadlockError) |
| 32–33 | `WORKFLOW_TASK_SCHEDULED / STARTED` |
| 34 | `WORKFLOW_TASK_TIMED_OUT` ← repeated indefinitely |

---

## Impact

- Affected workflows are **permanently stuck** — they never reach a terminal state
  (completed or failed) without manual operator intervention.
- The Juju unit shows **false green** status, providing no insight.
- The worker **slot is permanently occupied**, reducing available concurrency for all other workflows on the same worker.

### To Reproduce

1. Deploy `temporal-worker-k8s-operator` with a worker image whose application runs a blocking call for more than a few seconds
3. Observe that the unit shows **active/idle** despite the worker logging workflow eviction in a loop.

### Environment

Juju controller: 3.5.5

### Relevant log output

```shell
2026-05-25T07:31:53.657Z [temporal-worker]
  ERROR:temporalio.worker._workflow:Failed handling activation on workflow
  with run ID f04e62c4-abc4-4b97-8d68-7c4b2ae45a09
  temporalio.worker._workflow._DeadlockError: [TMPRL1101] Potential deadlock
  detected: workflow didn't yield within 2 second(s).
  ...
  File "//app/boarding/workflows/workflow_helpers.py", line 275, in process_plugin_result
    workflow_logger.info(f"{plugin_value} {lifecycle} completed successfully", ...)
2026-05-25T07:31:54.647Z [temporal-worker]
  ERROR:temporalio.worker._workflow:Failed running eviction job for run ID
  f04e62c4-abc4-4b97-8d68-7c4b2ae45a09, continually retrying eviction. Since
  eviction could not be processed, this worker may not complete and the slot
  may remain forever used unless it eventually completes.
  RuntimeError: Eviction processed, but 1 tasks remain.
```

## References

- Temporal SDK deadlock detection: https://docs.temporal.io/develop/python/debugging#deadlock-detection
- Pebble log API: https://documentation.ubuntu.com/pebble/en/latest/reference/api-access/
- Python `logging.handlers.QueueHandler`: https://docs.python.org/3/library/logging.handlers.html#queuehandler

---

## Recovery for Stuck Workflows

1. Deploy the fixed worker image.
2. The pod restarts, all in-memory workflow cache slots are cleared.
3. Temporal's sticky task queue timeout (~5 s of worker absence) expires.
4. Temporal routes the stuck workflow tasks to the fresh worker via the normal task queue.
5. The worker replays history cleanly using fixed async code.
6. The workflows continue and complete normally.

> If the workflow task timeout (`workflow_task_timeout`) has already elapsed
> and Temporal has marked the workflow as failed, use **Reset Workflow** in the
> Temporal UI to replay from the last completed event.

## Bug fix

A PR will be posted shortly with a suggested resolution for this bug report.

kot0dama:
## Description

Depends on #105 (must merge first). Once #105 merges, this PR will show only the eviction-loop changes.

When the Temporal Python SDK fails to evict a cached workflow instance (because a blocking call inside the workflow coroutine caused a _DeadlockError and left a pending asyncio task), it logs at ERROR level:

> Failed running eviction job for run ID <X>, continually retrying eviction.
> Since eviction could not be processed, this worker may not complete and the
>  slot may remain forever used unless it eventually completes.

The worker process continues running (Pebble service status: ACTIVE), so the existing service-running check does not catch this condition. The Juju unit falsely reports active/idle while the affected slot is permanently occupied and the stuck workflow tasks cycle between `WORKFLOW_TASK_FAILED` and `WORKFLOW_TASK_TIMED_OUT` indefinitely.

While the root cause of the eviction loop lies within the code and a fix is required, the charm could detect such situations and give an immediate hint to the user checking on `juju status`.

This adds `_has_eviction_loop_error(container)` which uses the Pebble raw log API (`GET /v1/logs?services=temporal-worker&n=200`) to scan the last 200 log lines for the sentinel phrase `continually retrying eviction`. Called in `_on_update_status` after the service-running check; sets `BlockedStatus` so the operator sees an actionable message in juju status:

  `temporal-worker: workflow eviction loop detected - restart required`

Recovery after redeploying the fix: the worker pod restarts, clearing all in-memory cache slots. Temporal's sticky task queue timeout expires and routes stuck workflow tasks to the fresh worker, which replays cleanly.

No manual Temporal UI intervention is required in the common case.

Also adds:
- test_eviction_loop_detected_in_pebble_logs scenario test
- test_no_eviction_loop_sets_active scenario test

This fixes #106. Of course, feel free to suggest a different resolution path!

---
## Checklist (all that apply)
- [x] Unit tests added or updated
- [ ] Documentation updated

### DIFF
--- a/src/charm.py
+++ b/src/charm.py
@@ -195,6 +195,19 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             self._update(event)
             return
 
+        try:
+            service = container.get_service(self.name)
+            if not service.is_running():
+                logger.error(
+                    "temporal-worker service is not running; it may be caught in a crash/restart loop - check logs for details"
+                )
+                self.unit.status = BlockedStatus(
+                    "temporal-worker service is not running; check logs for crash details"
+                )
+                return
+        except pebble.APIError as e:
+            logger.warning(f"Could not retrieve service status: {e}")
+
         self.unit.status = ActiveStatus(
             f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
         )
--- a/src/charm.py
+++ b/src/charm.py
@@ -208,6 +208,17 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
         except pebble.APIError as e:
             logger.warning(f"Could not retrieve service status: {e}")
 
+        if self._has_eviction_loop_error(container):
+            logger.error(
+                "Temporal SDK workflow eviction loop detected in pebble logs: "
+                "the worker has stuck slots and may not complete workflow tasks. "
+                "Restart the worker to recover (juju run temporal-worker-k8s/<unit> restart)."
+            )
+            self.unit.status = BlockedStatus(
+                "temporal-worker: workflow eviction loop detected - restart required"
+            )
+            return
+
         self.unit.status = ActiveStatus(
             f"worker listening to namespace {self.config['namespace']!r} on queue {self.config['queue']!r}"
         )
@@ -225,6 +236,39 @@ class TemporalWorkerK8SOperatorCharm(CharmBase):
             plan = container.get_plan().to_dict()
             return bool(plan and plan["services"].get(self.name, {}))
         except pebble.ConnectionError:
+            return False
+
+    def _has_eviction_loop_error(self, container) -> bool:
+        """Scan recent pebble service logs for the Temporal SDK workflow eviction loop error.
+
+        When a workflow task deadlocks (e.g. due to a blocking call inside the workflow
+        coroutine) the Temporal SDK fails to cleanly evict the cached workflow state and
+        logs the following at ERROR level:
+
+          "Failed running eviction job for run ID <X>, continually retrying eviction.
+           Since eviction could not be processed, this worker may not complete and the
+           slot may remain forever used unless it eventually completes."
+
+        The service continues running (pebble status stays ACTIVE), so this condition
+        is invisible to a simple service-status check.  This method reads the last 200
+        log lines from the pebble log API and checks for the sentinel phrase.
+
+        Args:
+            container: application container
+
+        Returns:
+            True if the eviction loop error was found in recent service logs.
+        """
+        try:
+            response = container._pebble._request_raw(
+                "GET",
+                "/v1/logs",
+                query={"services": self.name, "n": "200"},
+            )
+            log_data = response.read().decode("utf-8", errors="replace")
+            return "continually retrying eviction" in log_data
+        except Exception as e:
+            logger.debug(f"Could not scan pebble logs for eviction error: {e}")
             return False
 
     def get_auth_config_from_juju_secret(self) -> dict:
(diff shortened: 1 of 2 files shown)


DnPlas:
Closing this PR in favour of https://github.com/canonical/temporal-worker-k8s-operator/pull/110, which is already merged. Thanks @kot0dama !

kot0dama:
@DnPlas agreed, thanks for handling it while I was away!

PR #110, opened 2026-06-26
feat(charm): add pebble check for eviction loops

### LINKED ISSUE #106: Workflow Eviction Loop Not Surfaced in Charm Status

DnPlas:
## Description

Adds eviction-loop-check pebble check and blocks the unit whenever it fires
since the fix for an eviction loop is in the workflow code.
This commit also adds tests for the new functionality.

Fixes #106

---
## Testing instructions

1. Build and deploy the Charmed Temporal solution and this worker charm
2. Include a workflow that causes an eviction loop, in Python for instance:

```
  import asyncio
  from temporalio import workflow

  @workflow.defn
  class EvictionLoopWorkflow:
      @workflow.run
      async def run(self) -> None:
          while True:
              try:
                  await workflow.sleep(3600)
              except BaseException:
                  # Swallows asyncio.CancelledError, which the SDK uses internally
                  # to evict the workflow from the worker cache. The SDK retries the
                  # eviction indefinitely, logging "continually retrying eviction".
                  pass
```
3. Wait for the worker charm to be blocked with the message "eviction loop detected - fix workflow code before restarting" 

There was a suggestion from @shayancanonical to use the DB to query the logs instead of Pebble logs, after consideration, I decided to move forward with pebble logs since:

1. The eviction loop is purely a worker-side, in-memory condition
2. The Temporal server DB is not the worker's DB
3. The log is the earliest and most precise signal - as soon as the SDK sends the signal, we'll catch it

---
## Checklist (all that apply)
- [ ] Documentation updated

### DIFF
--- a/tests/scenario/test_charm.py
+++ b/tests/scenario/test_charm.py
@@ -682,7 +682,7 @@ def test_eviction_loop_check_detected(context, state, temporal_worker_container)
         state_out = context.run(context.on.pebble_check_failed(container, check_info), state_out)
 
     assert state_out.unit_status == ops.BlockedStatus(
-        "temporal-worker: eviction loop detected - fix workflow code before restarting"
+        "eviction loop detected - fix workflow code before restarting"
     )
 
 
--- a/tests/scenario/test_charm.py
+++ b/tests/scenario/test_charm.py
@@ -622,18 +622,18 @@ def test_blocked_by_missing_db_name(context, state, temporal_worker_container, c
     assert state_out.unit_status == ops.BlockedStatus("Invalid config: db name value missing")
 
 
-def _make_check_state(state_out):
-    """Return (container, check_info, updated_state) with the pebble check registered.
+def _make_check_state(state_out, name, threshold):
+    """Return (container, check_info, updated_state) with the named check registered.
 
     See https://github.com/canonical/operator/issues/2565: the consistency checker builds
     all_checks using raw container names but compares against the normalised event name,
     so containers with hyphens always fail. Patching here until the bug is fixed.
     """
     check_info = ops.testing.CheckInfo(
-        name="start-worker-check",
+        name=name,
         level=ops.pebble.CheckLevel.UNSET,
         startup=ops.pebble.CheckStartup.UNSET,
-        threshold=3,
+        threshold=threshold,
     )
     container = dataclasses.replace(
         state_out.get_container("temporal-worker"),
@@ -646,7 +646,7 @@ def test_pebble_check_failed(context, state, temporal_worker_container):
     state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
     state_out = context.run(context.on.config_changed(), state_out)
 
-    container, check_info, state_out = _make_check_state(state_out)
+    container, check_info, state_out = _make_check_state(state_out, "start-worker-check", threshold=3)
 
     # ops-scenario 7.21.1 bug: consistency checker does not normalise container names
     with unittest.mock.patch("scenario._consistency_checker.check_consistency"):
@@ -659,7 +659,7 @@ def test_pebble_check_recovered(context, state, temporal_worker_container, names
     state_out = context.run(context.on.pebble_ready(temporal_worker_container), state)
     state_out = context.run(context.on.config_changed(), state_out)
 
-    container, check_info, state_out = _make_check_state(state_out)
+    container, check_info, state_out = _make_check_state(state_out, "start-worker-check", threshold=3)
 
     # ops-scenario 7.21.1 bug: consistency checker does not normalise container names
     with unittest.mock.patch("scenario._consistency_checker.check_consistency"):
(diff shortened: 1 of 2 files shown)


shayancanonical (dismissed):
looks good, still curious if there's a better way to detect the eviction loop compared to tailing and scanning logs

tests/scenario/test_charm.py
```diff
@@ -671,6 +671,21 @@ def test_pebble_check_recovered(context, state, temporal_worker_container, names
...
+    state_out = context.run(context.on.config_changed(), state_out)
+
+    container, check_info, state_out = _make_check_state(state_out, "eviction-loop-check", threshold=1)
+
+    # ops-scenario 7.21.1 bug: consistency checker does not normalise container names
+    with unittest.mock.patch("scenario._consistency_checker.check_consistency"):
+        state_out = context.run(context.on.pebble_check_failed(container, check_info), state_out)
+
+    assert state_out.unit_status == ops.BlockedStatus(
+        "temporal-worker: eviction loop detected - fix workflow code before restarting"
```
nikita-jare: This assertion has a temporal-worker: prefix, but the [handler](https://github.com/canonical/temporal-worker-k8s-operator/pull/110/changes#diff-b9ed39bbc9c0387bd3e07da31d13373745534a1cd723d3e292c73496b12e307cR215) sets the status without it.
Dropping the prefix (or adding the suffix to the handler) should fix it

nikita-jare (commented):
LGTM except for this tiny statement change

tests/scenario/test_charm.py
DnPlas: good catch, I decided midway that there was no reason to add the app name in the status message, but forgot to edit the test. Anyway, fixed in bfccad3.

nikita-jare (approved):
Thank you @DnPlas

shayancanonical (approved):

